PR #66, opened 2018-09-08
Fixed < key in keyboard not working

jhakonen:
This fixes less-than / greater-than / pipe -key (see picture) in my keyboard not doing anything while streaming.

photo_2018-09-08_14-21-16


jhakonen:
Thank you! Now I can do programming through Moonlight, since that key was rather crucial for writing conditions. 😄

PR #67, opened 2018-09-08
Capture mouse cursor when clicking inactive window

jhakonen:
Currently the mouse capture doesn't trigger until Moonlight's window has been activated and user has clicked the window's client area. I think this is fine when you use e.g. ALT+TAB to activate Moonlight and then click the window to enable mouse capture.
However, if you use mouse only to do this, then you need to click the window twice, once to activate it, and second time to enable mouse capture. I found that a bit unintuitive to use.
This pull request changes that, now only requiring a single click to both activate the window and toggle mouse capture.

### DIFF
--- a/app/streaming/session.cpp
+++ b/app/streaming/session.cpp
@@ -929,6 +929,26 @@ void Session::exec(int displayOriginX, int displayOriginY)
         }
 
         case SDL_WINDOWEVENT:
+            // Capture mouse cursor when user actives the window by clicking on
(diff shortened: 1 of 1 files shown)


cgutman:
This seems to work as intended on Windows and macOS, but on Linux (Fedora 28) it results in a click event sent to the host PC, since apparently the focus gain is delivered before the click event. It appears the fix isn't necessary on Linux, so I'm going to guard it with `#if` to only run on Windows and Mac.

Issue #30, opened 2018-08-08
Feature Request - Command line launch options


bellinger61:
Linux command line support, no documentation available on this currently and previous java based command line options do not function with recommended run commands associated with flatpak

cgutman:
There's no command line option docs because there are no supported command line options yet. This client is a complete rewrite sharing none of the previous Java client code, so none of the previous options will work on this client until we implement them.

bellinger61:
Thank you for the quick response, I've been a long time user of moonlight on many platforms but have been stuck using steam streaming on my repurposed Mac mini retropie linux box, this moonlight client already far surpasses the quality I have gotten from the steam client hands down.

Really appreciate your and the rest of the teams work and look forward to the Linux and PC versions continued development!!!

jhakonen:
I would like to have this feature too. It would be great to have a handy shortcut on desktop to open a game directly in my gaming PC. So, I've been hacking the code and trying to make this feature happen. I have it mostly done. I can open a game stream directly from command line and it works. But, it's very sunny case, no error handling whatsoever. So, still needs work.

Launching the stream shows first a segue window, like the one you get after clicking a game in the actual menu UI, which shows connection progress.

Here's what I have come up with the command line options:
```
> moonlight.exe --help
Usage: moonlight.exe [options] action

Options:
  -?, -h, --help  Displays this help.
  -v, --version   Displays version information.

Arguments:
  action          Action to execute
```
If you don't provide the action, the normal moonlight UI will open.
Here's help for starting game stream:
```
> moonlight.exe stream --help
Usage: moonlight.exe [options] stream host game

Options:
  -?, -h, --help                   Displays this help.
  -v, --version                    Displays version information.
  --720                            Use 1280x720 resolution.
  --1080                           Use 1920x1080 resolution.
  --1440                           Use 2560x1440 resolution.
  --4K                             Use 3840x2160 resolution.
  --width <width>                  Specify horizontal resolution to use.
  --height <height>                Specify vertical resolution to use.
  --vsync                          Use V-Sync.
  --no-vsync                       Do not use V-Sync.
  --fps <fps>                      Specify FPS to use.
  --bitrate <bitrate>              Specify bitrate in Kbps to use.
  --display <display>              Select display mode:
                                   borderless/fullscreen/windowed.
  --audio-config <audio-config>    Select audio config: auto/stereo/surround.
  --multi-controller               Use multiple controller support.
[...]

PR #74, opened 2018-09-12
Fix a possible crash with SDL cleanup on process exit

jhakonen:
While hacking the code, trying to to do something about #30, I encountered in rare occasions a crash on process exit. The crash occurs in SDL code within SDL_RunAudio() function in SDL_audio.c. Seems that the Moonlight called SDL_Quit() in main.cpp while SDL's audio thread was still running. 
The crash can be easily reproduced using by debug build and running Moonlight in debugger:

1. Put a breakpoint to SDL_Quit() -call in main.
2. Then, through Moonlight start a game, and when game is running, press ALT+F4 twice fast. This closes both the SDL window and the main menu window almost simultaneusly.
3. Execution pauses at the breakpoint. Continue execution.
4. Process crashes within SDL.

This fix gives time for the session cleanup to properly do the SDL cleanup, before letting execution leave the event loop, and reach SDL_Quit() call.

**EDIT:** I used QtCreator with x64 MSVC 2017 (community edition) compiler.

### DIFF
--- a/app/streaming/session.cpp
+++ b/app/streaming/session.cpp
@@ -1119,5 +1119,13 @@ DispatchDeferredCleanup:
     // When it is complete, it will release our s_ActiveSessionSemaphore
     // reference.
     QThreadPool::globalInstance()->start(new DeferredSessionCleanupTask());
+
+    // Prevents calling of SDL_Quit() in main while cleanup is running by
+    // giving the cleanup time to finish. May occur when all process's windows
+    // are closed at same time.
+    connect(QCoreApplication::instance(), &QCoreApplication::aboutToQuit, []() {
+        Session::s_ActiveSessionSemaphore.tryAcquire(1, 10000);
+    });
+
 }
 
--- a/app/streaming/session.cpp
+++ b/app/streaming/session.cpp
@@ -1124,7 +1124,9 @@ DispatchDeferredCleanup:
     // giving the cleanup time to finish. May occur when all process's windows
     // are closed at same time.
     connect(QCoreApplication::instance(), &QCoreApplication::aboutToQuit, []() {
-        Session::s_ActiveSessionSemaphore.tryAcquire(1, 10000);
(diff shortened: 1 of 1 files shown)


jhakonen:
Did the second commit, as I realized a bit too late that the code actually does one connect() -per session. If you open/close sessions several times within same process, the callbacks pile up. When the event loop exits it calls the lambda callback as many times as there were sessions. So, X sessions could have led to X*10s wait time on exit.

app/streaming/session.cpp
```diff
@@ -1119,5 +1119,15 @@ void Session::exec(int displayOriginX, int displayOriginY)
     // When it is complete, it will release our s_ActiveSessionSemaphore
     // reference.
     QThreadPool::globalInstance()->start(new DeferredSessionCleanupTask());
+
+    // Prevents calling of SDL_Quit() in main while cleanup is running by
+    // giving the cleanup time to finish. May occur when all process's windows
+    // are closed at same time.
+    connect(QCoreApplication::instance(), &QCoreApplication::aboutToQuit, []() {
+        if (Session::s_ActiveSessionSemaphore.tryAcquire(1, 10000)) {
```
cgutman: Using `tryAcquire()` means that it's still possible to crash this way if connection cleanup takes longer than 10 seconds.

I don't really like this solution though, because it's a workaround for the real problem which is that we can leave `Session::exec()` with SDL threads still alive (since the `SDL_CloseAudioDevice()` call was taking place as a result of `LiStopConnection()` which happens asynchronously).

I moved SDL audio cleanup onto the main thread in https://github.com/moonlight-stream/moonlight-qt/commit/64bc189010b6550e7da90307a065157a0ecbda89 for unrelated reasons, but it should solve this issue too.

jhakonen: Yeah, seems to be fixed indeed 👍

PR #89, opened 2018-09-26
Add command line parameters. Fixes #30

### LINKED ISSUE #30: Feature Request - Command line launch options

jhakonen:
Final outputs for --help:
```
> moonlight --help
Usage: moonlight [options] <action>

Starts Moonlight normally if no arguments are given.

Available actions:
  stream          Start streaming a game

See 'moonlight <action> --help' for help of specific action.

Options:
  -?, -h, --help  Displays this help.
  -v, --version   Displays version information.

Arguments:
  action          Action to execute
```
```
> moonlight stream --help
Usage: moonlight [options] stream <host> <game>

Starts directly streaming a given game.

Options:
  -?, -h, --help                   Displays this help.
  -v, --version                    Displays version information.
  --720                            Use 1280x720 resolution.
  --1080                           Use 1920x1080 resolution.
  --1440                           Use 2560x1440 resolution.
  --4K                             Use 3840x2160 resolution.
  --resolution <resolution>        Specify custom <width>x<height> resolution
                                   to use.
  --vsync                          Use V-Sync.
  --no-vsync                       Do not use V-Sync.
  --fps <fps>                      Specify FPS to use.
  --bitrate <bitrate>              Specify bitrate in Kbps to use.
  --display-mode <display-mode>    Select display mode:
                                   borderless/fullscreen/windowed.
  --audio-config <audio-config>    Select audio config: auto/stereo/surround.
  --multi-controller               Use multiple controller support.
  --no-multi-controller            Do not use multiple controller support.
  --mouse-acceleration             Use mouse acceleration.
  --no-mouse-acceleration          Do not use mouse acceleration.
  --game-optimization              Use game optimizations.
  --no-game-optimization           Do not use game optimizations.
  --audio-on-host                  Use audio on host PC.
  --no-audio-on-host               Do not use audio on host PC.
  --video-codec <video-codec>      Select video codec: H.264/HEVC/auto.
  --video-decoder <video-decoder>  Select video decoder:
                                   auto/hardware/software.

Arguments:
  stream                           Start stream
  host                             Host computer name, uuid or address
  game                             Game name
```

To start streaming a game do e.g.:
`> moonlight MyAwesomeGamingPC "XCOM 2" --1080 --display-mode windowed `

When the game is launched:
1. main() creates main.qml window and StackView's initialView is set to CliStartStreamSeque.qml
2. CliStartStreamSeque.qml executes CliStartStream::Launcher
3. Launcher waits until PC and the game is found and then creates a Session
4. CliStartStreamSeque.qml adds StreamSeque.qml to StackView and passes the Session to it, which then handles the actual Session execution
5. Once Session ends, the process quits

### DIFF
--- a/app/cli/commandlineparser.cpp
+++ b/app/cli/commandlineparser.cpp
@@ -1,0 +1,335 @@
+#include "commandlineparser.h"
+
+#include <QCommandLineParser>
+#include <QRegularExpression>
+
+#if defined(Q_OS_WIN)
+#include <qt_windows.h>
+#endif
+
+static bool inRange(int value, int min, int max)
+{
+    return value >= min && value <= max;
+}
+
+class CommandLineParser : public QCommandLineParser
+{
+public:
+    enum MessageType {
+        Info,
+        Error
+    };
+
+    void setupCommonOptions()
+    {
+        setSingleDashWordOptionMode(QCommandLineParser::ParseAsLongOptions);
+        addHelpOption();
+        addVersionOption();
+    }
+
+    void handleHelpAndVersionOptions()
+    {
+        if (isSet("help")) {
+            showInfo(helpText());
+        }
+        if (isSet("version")) {
+            showVersion();
+        }
+    }
+
+    void showMessage(QString message, MessageType type) const
+    {
+    #if defined(Q_OS_WIN)
+        UINT flags = MB_OK | MB_TOPMOST | MB_SETFOREGROUND;
+        flags |= (type == Info ? MB_ICONINFORMATION : MB_ICONERROR);
+        QString title = "Moonlight";
+        MessageBoxW(nullptr, reinterpret_cast<const wchar_t *>(message.utf16()),
+                    reinterpret_cast<const wchar_t *>(title.utf16()), flags);
+    #endif
+        fputs(qPrintable(message), type == Info ? stdout : stderr);
+    }
+
+    [[ noreturn ]] void showInfo(QString message) const
+    {
+        showMessage(message, Info);
+        exit(0);
+    }
+
+    [[ noreturn ]] void showError(QString message) const
+    {
+        showMessage(message, Error);
+        exit(1);
+    }
+
+    int getIntOption(QString name) const
+    {
+        bool ok;
+        int intValue = value(name).toInt(&ok);
+        if (!ok) {
+            showError(QString("Invalid %1 value: %2").arg(name, value(name)));
+        }
+        return intValue;
+    }
+
+    bool getToggleOptionValue(QString name, bool defaultValue) const
+    {
+        QRegularExpression re(QString("^(%1|no-%1)$").arg(name));
+        QStringList options = optionNames().filter(re);
+        if (options.isEmpty()) {
+            return defaultValue;
+        } else {
+            return options.last() == name;
+        }
+    }
+
+    QString getChoiceOptionValue(QString name) const
+    {
+        if(!m_Choices[name].contains(value(name))) {
+            showError(QString("Invalid %1 choice: %2").arg(name, value(name)));
+        }
+        return value(name);
+    }
+
+    QPair<int,int> getResolutionOptionValue(QString name) const
+    {
+        QRegularExpression re("^(\\d+)x(\\d+)$", QRegularExpression::CaseInsensitiveOption);
+        auto match = re.match(value(name));
+        if (!match.hasMatch()) {
+            showError(QString("Invalid %1 format: %2").arg(name, value(name)));
+        }
+        return qMakePair(match.captured(1).toInt(), match.captured(2).toInt());
+    }
+
+    void addFlagOption(QString name, QString descriptiveName)
+    {
+        addOption(QCommandLineOption(name, QString("Use %1.").arg(descriptiveName)));
+    }
+
+    void addToggleOption(QString name, QString descriptiveName)
+    {
+        addOption(QCommandLineOption(name, QString("Use %1.").arg(descriptiveName)));
+        addOption(QCommandLineOption("no-" + name, QString("Do not use %1.").arg(descriptiveName)));
+    }
+
+    void addValueOption(QString name, QString descriptiveName)
+    {
+        addOption(QCommandLineOption(name, QString("Specify %1 to use.").arg(descriptiveName), name));
+    }
+
+    void addChoiceOption(QString name, QString descriptiveName, QStringList choices)
+    {
+        addOption(QCommandLineOption(name, QString("Select %1: %2.").arg(descriptiveName, choices.join('/')), name));
+        m_Choices[name] = choices;
+    }
+
+private:
+    QMap<QString, QStringList> m_Choices;
+};
+
+GlobalCommandLineParser::GlobalCommandLineParser()
+{
+}
+
+GlobalCommandLineParser::~GlobalCommandLineParser()
+{
+}
+
+GlobalCommandLineParser::ParseResult GlobalCommandLineParser::parse(const QStringList &args)
+{
+    CommandLineParser parser;
+    parser.setupCommonOptions();
+    parser.setApplicationDescription(
+        "\n"
+        "Starts Moonlight normally if no arguments are given.\n"
+        "\n"
+        "Available actions:\n"
+        "  stream          Start streaming a game\n"
+        "\n"
+        "See 'moonlight <action> --help' for help of specific action."
+    );
+    parser.addPositionalArgument("action", "Action to execute", "<action>");
+    parser.parse(args);
+    auto posArgs = parser.positionalArguments();
+    QString action = posArgs.isEmpty() ? QString() : posArgs.first();
+
+    if (action == "") {
+        parser.handleHelpAndVersionOptions();
+        return NormalStartRequested;
+    } else if (action == "stream") {
+        return StreamRequested;
+    } else {
+        parser.showError(QString("Invalid action: %1").arg(action));
+    }
+}
+
+StreamCommandLineParser::StreamCommandLineParser()
+{
+    m_WindowModeMap = {
+        {"fullscreen", StreamingPreferences::WM_FULLSCREEN},
+        {"windowed",   StreamingPreferences::WM_WINDOWED},
+        {"borderless", StreamingPreferences::WM_FULLSCREEN_DESKTOP},
+    };
+    m_AudioConfigMap = {
+        {"auto",     StreamingPreferences::AC_AUTO},
+        {"stereo",   StreamingPreferences::AC_FORCE_STEREO},
+        {"surround", StreamingPreferences::AC_FORCE_SURROUND},
+    };
+    m_VideoCodecMap = {
+        {"auto",  StreamingPreferences::VCC_AUTO},
+        {"H.264", StreamingPreferences::VCC_FORCE_H264},
+        {"HEVC",  StreamingPreferences::VCC_FORCE_HEVC},
+    };
+    m_VideoDecoderMap = {
+        {"auto",     StreamingPreferences::VDS_AUTO},
+        {"software", StreamingPreferences::VDS_FORCE_HARDWARE},
+        {"hardware", StreamingPreferences::VDS_FORCE_SOFTWARE},
+    };
+}
+
+StreamCommandLineParser::~StreamCommandLineParser()
+{
+}
+
+void StreamCommandLineParser::parse(const QStringList &args, StreamingPreferences *preferences)
+{
+    CommandLineParser parser;
+    parser.setupCommonOptions();
+    parser.setApplicationDescription(
+        "\n"
+        "Starts directly streaming a given game."
+    );
+    parser.addPositionalArgument("stream", "Start stream");
+
+    // Add other arguments and options
+    parser.addPositionalArgument("host", "Host computer name, uuid or address", "<host>");
+    parser.addPositionalArgument("game", "Game name", "<game>");
+
+    parser.addFlagOption("720",  "1280x720 resolution");
+    parser.addFlagOption("1080", "1920x1080 resolution");
+    parser.addFlagOption("1440", "2560x1440 resolution");
+    parser.addFlagOption("4K", "3840x2160 resolution");
+    parser.addValueOption("resolution", "custom <width>x<height> resolution");
+    parser.addToggleOption("vsync", "V-Sync");
+    parser.addValueOption("fps", "FPS");
+    parser.addValueOption("bitrate", "bitrate in Kbps");
+    parser.addChoiceOption("display-mode", "display mode", m_WindowModeMap.keys());
+    parser.addChoiceOption("audio-config", "audio config", m_AudioConfigMap.keys());
+    parser.addToggleOption("multi-controller", "multiple controller support");
+    parser.addToggleOption("mouse-acceleration", "mouse acceleration");
+    parser.addToggleOption("game-optimization", "game optimizations");
+    parser.addToggleOption("audio-on-host", "audio on host PC");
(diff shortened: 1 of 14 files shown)


jhakonen:
Oh, both Mac and Linux compilation fails.
Mac's issue is obvious. But, Linux's isn't as so. I'll try to get Ubuntu running in VM to figure out how to fix the issue there.

jhakonen:
There, errors fixed and all builds seem to pass (if you don't count appveyor's internal server errors).

app/cli/commandlineparser.cpp
```diff
@@ -0,0 +1,335 @@
...
+        "\n"
+        "See 'moonlight <action> --help' for help of specific action."
+    );
+    parser.addPositionalArgument("action", "Action to execute", "<action>");
+    parser.parse(args);
+    auto posArgs = parser.positionalArguments();
+    QString action = posArgs.isEmpty() ? QString() : posArgs.first();
+
+    if (action == "") {
+        parser.handleHelpAndVersionOptions();
```
cgutman: Can you put a comment here stating that this won't return if `--help` or `--version` is specified?

app/cli/commandlineparser.cpp
```diff
@@ -0,0 +1,335 @@
...
+#if defined(Q_OS_WIN)
+#include <qt_windows.h>
+#endif
+
+static bool inRange(int value, int min, int max)
+{
+    return value >= min && value <= max;
+}
+
+class CommandLineParser : public QCommandLineParser
```
cgutman: Can you making the argument parsing in this class case-insensitive?

app/cli/startstream.cpp
```diff
@@ -0,0 +1,190 @@
...
+        NvApp app;
+
+        switch (event.type) {
+        case Event::Executed:
+            if (m_State == StateInit) {
+                m_State = StateSeekComputer;
+                m_TimeoutTimer->start(COMPUTER_SEEK_TIMEOUT);
+                m_ComputerManager = event.computerManager;
+                q->connect(m_ComputerManager, &ComputerManager::computerStateChanged,
+                           q, &Launcher::onComputerUpdated);
```
cgutman: Connecting `ComputerManager::computerAddCompleted()` would allow you to catch an unreachable host before the 15 second timeout

app/gui/CliStartStreamSegue.qml
```diff
@@ -0,0 +1,78 @@
...
+import QtQuick.Controls 2.2
+import QtQuick.Dialogs 1.2
+
+import ComputerManager 1.0
+
+Item {
+    visible: false
+
+    function onSearchingComputer(name) {
+        stageLabel.text = "Searching computer " + name + "..."
```
cgutman: I think we could just use a generic message here like: `"Establishing connection to PC..."`

app/gui/CliStartStreamSegue.qml
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    function onSearchingApp(name) {
+        stageLabel.text = "Searching application " + name + "..."
```
cgutman: Similar to above, I think we could just use `"Loading app list..."`

app/gui/CliStartStreamSegue.qml
```diff
@@ -0,0 +1,78 @@
...
+        id: errorDialog
+        modality:Qt.WindowModal
+        icon: StandardIcon.Critical
+        standardButtons: StandardButton.Ok | StandardButton.Help
+        onHelp: {
+            Qt.openUrlExternally("https://github.com/moonlight-stream/moonlight-docs/wiki/Troubleshooting");
+        }
+        onVisibleChanged: {
+            if (!visible) {
+                Qt.quit();
```
cgutman: This is to quit when the dialog is closed?

app/gui/StreamSegue.qml
```diff
@@ -72,11 +73,15 @@ Item {
             // Run the streaming session to completion
             session.exec(Screen.virtualX, Screen.virtualY)
```
cgutman: This is going to be slightly problematic for multi-monitor configurations, since the user won't have the ability to position the UI on their preferred display. Maybe we should add an option to specify the launch monitor?

I don't think this needs to be done before merging though.

app/streaming/session.cpp
```diff
@@ -275,8 +275,9 @@ int Session::getDecoderCapabilities(StreamingPreferences::VideoDecoderSelection
     return caps;
 }
 
-Session::Session(NvComputer* computer, NvApp& app)
-    : m_Computer(computer),
+Session::Session(NvComputer* computer, NvApp& app, StreamingPreferences *preferences)
+    : m_Preferences(preferences ? preferences : new StreamingPreferences(this)),
```
cgutman: We need to delete `m_Preferences` in the destructor if `preferences == nullptr`

app/main.cpp
```diff
@@ -299,8 +302,31 @@ int main(int argc, char *argv[])
...
+
+    GlobalCommandLineParser parser;
+    switch (parser.parse(app.arguments())) {
+    case GlobalCommandLineParser::NormalStartRequested:
+        initialView = "PcView.qml";
+        break;
+    case GlobalCommandLineParser::StreamRequested:
+        {
+            initialView = "CliStartStreamSegue.qml";
+            StreamingPreferences* preferences = new StreamingPreferences(&app);
```
cgutman: This StreamingPreferences object needs to be deleted at some point.

app/cli/commandlineparser.cpp
jhakonen: Positional arguments should be trivial to handle in case insensitive manner. Same for option values.

However, QCommandLineParser doesn't seem to provide any easy to way to parse option names in case insensitive manner. But perhaps unknownOptionNames() could be used to detect any options that were given in wrong case. And then adjust them to correct case and reparse them... Will have to try.

app/cli/commandlineparser.cpp
```diff
@@ -0,0 +1,335 @@
...
+        MessageBoxW(nullptr, reinterpret_cast<const wchar_t *>(message.utf16()),
+                    reinterpret_cast<const wchar_t *>(title.utf16()), flags);
+    #endif
+        fputs(qPrintable(message), type == Info ? stdout : stderr);
+    }
+
+    [[ noreturn ]] void showInfo(QString message) const
+    {
+        showMessage(message, Info);
+        exit(0);
```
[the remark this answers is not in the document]
jhakonen: But, wouldn't that just try to exit from the event loop? Event loop which isn't running at this point.

I could write this without calling `exit()`, and just return to `main()`, but then the `showVersion()` would need to be shown differently as it calls the global `exit()` internally. Otherwise the message showing would behave inconsistently.

app/cli/commandlineparser.cpp
```diff
@@ -0,0 +1,335 @@
...
+        if (options.isEmpty()) {
+            return defaultValue;
+        } else {
+            return options.last() == name;
+        }
+    }
+
+    QString getChoiceOptionValue(QString name) const
+    {
+        if(!m_Choices[name].contains(value(name))) {
```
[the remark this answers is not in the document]
jhakonen: 👍

app/cli/commandlineparser.cpp
jhakonen: Can do

app/cli/startstream.cpp
jhakonen: Umm, there is also the mDNS polling going on. I assume that the `computerAddCompleted()` would not tell if mDNS can't find it, right?

app/gui/CliStartStreamSegue.qml
jhakonen: Ok 👍

app/gui/CliStartStreamSegue.qml
jhakonen: Ok

app/gui/CliStartStreamSegue.qml
jhakonen: Yep. Had to do it like this to get notified when user clicks x-button in the title bar. There doesn't seem to be any other signal that would trigger for it.

app/gui/StreamSegue.qml
jhakonen: Yeah, that sound good idea to me.

app/main.cpp
jhakonen: Note that I'm passing the `QCoreApplication` as a parent to the `StreamingPreferences`. That handles deletion of the preferences object automatically when `QCoreApplication` object gets deleted.

app/streaming/session.cpp
jhakonen: As mentioned above, `Session` is set as parent of the `StreamingPreferences` (passing `this` as argument). This makes the `Session` object take ownership of the preferences object, and the preferences object gets deleted once `Session` is deleted.

app/cli/startstream.cpp
cgutman: It would only be emitted if the manual addComputer() fails. However, it now occurs to me that doing that would break resolving UUIDs (since add will fall for those).

I still don't like having to wait 15 seconds when the user mistypes an IP address or hostname. Maybe you could lower the timeouts from 15 to 10 seconds as a compromise?

app/cli/commandlineparser.cpp
jhakonen: Ok

app/cli/startstream.cpp
jhakonen: Yeah, that is fine for me. When I tested this, both the manual `addComputer()` and the mDNS quieries found the host computer almost instantiously. But, I was unsure if my experience was normal or not. So, I just picked some high-ish enough numbers as timeout values.

jhakonen: I'll lower the app list loading timeout to 10 too.

app/cli/startstream.cpp
```diff
@@ -0,0 +1,190 @@
...
+                if (matchComputer(event.computer) && isOnline(event.computer)) {
+                    if (isPaired(event.computer)) {
+                        m_State = StateSeekApp;
+                        m_TimeoutTimer->start(APP_SEEK_TIMEOUT);
+                        m_Computer = event.computer;
+                        m_ComputerManager->stopPollingAsync();
+                        emit q->searchingApp();
+                    } else {
+                        m_State = StateFailure;
+                        emit q->failed(QString("Computer %1 has not been paired").arg(m_ComputerName));
```
cgutman: Use `event.computer->Name` here instead of `m_ComputerName`. If I try to connect to a host by IP address or UUID that is not paired, it will give me back the IP address or UUID in the error message rather than the PC's friendly name.

cgutman: Also let's add a bit of extra info to that message like `"Computer %1 has not been paired. Please open Moonlight to pair before streaming."`

app/cli/startstream.cpp
```diff
@@ -0,0 +1,190 @@
...
+                    }
+                }
+            }
+            if (m_State == StateSeekApp) {
+                int index = getAppIndex();
+                if (-1 != index) {
+                    m_State = StateStartSession;
+                    m_TimeoutTimer->stop();
+                    app = m_Computer->appList[index];
+                    session = new Session(m_Computer, app, m_Preferences);
```
cgutman: We should do something in the case that the user is already streaming a different game (`m_Computer->currentGameId != 0 && m_Computer->currentGameId != app->id`). Otherwise, we'll hit the Session::exec() assert:
```
    // The UI should have ensured the old game was already quit
    // if we decide to stream a different game.
    Q_ASSERT(m_Computer->currentGameId == 0 ||
             m_Computer->currentGameId == m_App.id);
```

I would suggest an error message like `QString("%1 is already running. Please quit %1 to stream %2.").arg( runningApp.name).arg(app.name)`

app/cli/commandlineparser.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+    [[ noreturn ]] void showInfo(QString message) const
+    {
+        showMessage(message, Info);
+        exit(0);
+    }
+
+    [[ noreturn ]] void showError(QString message) const
+    {
+        showMessage(message, Error);
```
cgutman: Can we include the help text in this message? Maybe like `message + "\n\n" + helpText()`?

app/cli/commandlineparser.cpp
```diff
@@ -0,0 +1,363 @@
...
+    CommandLineParser parser;
+    parser.setupCommonOptions();
+    parser.setApplicationDescription(
+        "\n"
+        "Starts directly streaming a given game."
+    );
+    parser.addPositionalArgument("stream", "Start stream");
+
+    // Add other arguments and options
+    parser.addPositionalArgument("host", "Host computer name, uuid or address", "<host>");
```
cgutman: `"Host computer name, UUID, or IP address"`

app/cli/commandlineparser.cpp
```diff
@@ -0,0 +1,363 @@
...
+    parser.addPositionalArgument("game", "Game name", "<game>");
```
cgutman: let's put quotes around `<game>` to avoid confusion with users specifying games with spaces but not using quotes.

app/cli/startstream.cpp
jhakonen: 👍

app/cli/startstream.cpp
jhakonen: 👍

app/cli/commandlineparser.cpp
jhakonen: Can do

app/cli/commandlineparser.cpp
jhakonen: 👍

app/cli/commandlineparser.cpp
jhakonen: Done.
Which one you think is better here; game, app, or application? I'm not sure which one I should use. Goes also for text in `setApplicationDescription()`.

jhakonen: I think the text content in first argument `app` and third `"<app name>"` should remain same. So `app` and `"<app name>"` or `app` and third `"<app>"`. Otherwise there is a confusion between usage text and arguments list:
```
> moonlight stream --help
Usage: moonlight [options] stream <host> "<app name>"
...
Arguments:
...
  app                              App to stream
```

On the other hand I feel that having `app name` in the arguments list might be confusing, since white space has a meaning as a argument separator.

Thus I would keep the third parameter as `<app>`

cgutman: That sounds ok to me.

cgutman:
Thanks for doing this work.

app/gui/StreamSegue.qml
jhakonen: So, I tested this a bit. 

It seems that Qt doesn't offer a simple way to identify a display, like some integer 1, 2, ... etc. There's `QGuiApplication::screens()` which returns list of `QScreen`, but at least on my Win10 machine with two displays connected, both `QScreen`s return empty strings for `model()` and `serialNumber()`. `name()` returns `"\\\\.\\DISPLAY2"` and `"\\\\.\\DISPLAY1"`.
On Ubuntu running inside Virtualbox with one display, I get `"VGA-1"` as name. Model and serialNumber are empty there as well.

Perhaps this could work so that, there would a new command line action for listing displays, e,g,:
```
> moonlight.exe list-displays
\\.\DISPLAY2
\\.\DISPLAY1
```
And then you could pass the display name as an option to stream action:
```
> moonlight.exe stream MyAwesomePC "XCOM 2" --display \\.\DISPLAY1
```

PR #138, opened 2018-11-25
Add quit cli command and app quit option after stream session. Fixes #92

### LINKED ISSUE #92: Quit session from command line
I think it would really nice to have the option to close the current session from the command line.

We got command line parameters from #89, in the same way we could envision an action in addiction to 'stream' called 'quit' that closes the current session. This would be useful in a scenario where the main gui is not accessed often and sessions with a target application are launched from the command line: once launched a session, we currently can't really close it (which means we can't then launch a second, different session) without opening the main gui first and closing the existing session from there. Having a 'quit' action could allow for easy scripting e.g. close current existing session then launch application X.

Let me conclude by thanking every contributor for the great job so far, this client really is coming together as a nice piece of software!

jhakonen:
Adds `quit` CLI command:
```
$ moonlight quit --help
Usage: moonlight [options] quit <host>

Quit currently running app at given host.

Options:
  -?, -h, --help  Displays this help.
  -v, --version   Displays version information.

Arguments:
  quit            quit running app
  host            Host computer name, UUID, or IP address
```
Adds new options to `stream` CLI command:
```
$ moonlight stream --help
...
Options:
...
  --quit-after                     Use quit app after session.
  --no-quit-after                  Do not use quit app after session.
...
```
Those options override this new checkbox in UI preferences:
quit-app

In addition, when starting an app from CLI using `stream` command, the UI will now prompt user to quit previously running app if there is one, and it is not the same that user is trying to start:
stream-quit

### DIFF
--- a/app/backend/computermanager.h
+++ b/app/backend/computermanager.h
@@ -14,6 +14,8 @@
 #include <QSettings>
 #include <QRunnable>
 
+class Session;
+
 class MdnsPendingComputer : public QObject
 {
     Q_OBJECT
@@ -72,6 +74,7 @@ public:
     void pairHost(NvComputer* computer, QString pin);
 
     void quitRunningApp(NvComputer* computer);
+    Q_INVOKABLE void quitRunningApp(Session* session);
 
     QVector<NvComputer*> getComputers();
 
--- a/app/backend/nvhttp.cpp
+++ b/app/backend/nvhttp.cpp
@@ -57,8 +57,13 @@ NvHTTP::getCurrentGame(QString serverInfo)
     // GFE 2.8 started keeping currentgame set to the last game played. As a result, it no longer
     // has the semantics that its name would indicate. To contain the effects of this change as much
     // as possible, we'll force the current game to zero if the server isn't in a streaming session.
+    //
+    // However, current game info must be available also in other states than just _SERVER_BUSY as
+    // it is required for quitting currently running app. Quitting app occurs at end of stream if
+    // configured so. At that point the server state may be in some other state than _SERVER_BUSY
+    // for a short while, but that must not prevent quitting of the app.
     QString serverState = getXmlString(serverInfo, "state");
-    if (serverState != nullptr && serverState.endsWith("_SERVER_BUSY"))
+    if (serverState != nullptr && !serverState.endsWith("_SERVER_AVAILABLE"))
     {
         return getXmlString(serverInfo, "currentgame").toInt();
     }
--- a/app/cli/commandlineparser.cpp
+++ b/app/cli/commandlineparser.cpp
@@ -164,6 +164,7 @@ GlobalCommandLineParser::ParseResult GlobalCommandLineParser::parse(const QStrin
         "Starts Moonlight normally if no arguments are given.\n"
         "\n"
         "Available actions:\n"
+        "  quit            Quit currently running app\n"
         "  stream          Start streaming an app\n"
         "\n"
         "See 'moonlight <action> --help' for help of specific action."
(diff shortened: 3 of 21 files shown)


app/streaming/session.cpp
```diff
@@ -275,6 +275,16 @@ int Session::getDecoderCapabilities(StreamingPreferences::VideoDecoderSelection
...
 }
 
+NvComputer *Session::getComputer() const
+{
+    return m_Computer;
+}
+
+bool Session::shouldQuitAppAfter() const
+{
+    return m_Preferences->quitAppAfter;
```
cgutman: This should be exposed via StreamingPreferences itself, not the Session class.

app/backend/computermanager.h
```diff
@@ -72,6 +74,7 @@ class ComputerManager : public QObject
     void pairHost(NvComputer* computer, QString pin);
 
     void quitRunningApp(NvComputer* computer);
+    Q_INVOKABLE void quitRunningApp(Session* session);
```
cgutman: Why can't we use  `void quitRunningApp(NvComputer* computer)` instead?

app/backend/nvhttp.cpp
```diff
@@ -57,8 +57,13 @@ NvHTTP::getCurrentGame(QString serverInfo)
...
     // has the semantics that its name would indicate. To contain the effects of this change as much
     // as possible, we'll force the current game to zero if the server isn't in a streaming session.
+    //
+    // However, current game info must be available also in other states than just _SERVER_BUSY as
+    // it is required for quitting currently running app. Quitting app occurs at end of stream if
+    // configured so. At that point the server state may be in some other state than _SERVER_BUSY
+    // for a short while, but that must not prevent quitting of the app.
     QString serverState = getXmlString(serverInfo, "state");
-    if (serverState != nullptr && serverState.endsWith("_SERVER_BUSY"))
+    if (serverState != nullptr && !serverState.endsWith("_SERVER_AVAILABLE"))
```
cgutman: There were several versions of GFE that had a bug where they left the server in state `_SERVER_QUITTING` even after the game was closed, so we need to treat `_SERVER_QUITTING` as if it was `_SERVER_AVAILABLE`.

What state were you seeing the server in when you were seeing issues quitting?

app/cli/commandlineparser.cpp
```diff
@@ -164,6 +164,7 @@ GlobalCommandLineParser::ParseResult GlobalCommandLineParser::parse(const QStrin
         "Starts Moonlight normally if no arguments are given.\n"
         "\n"
         "Available actions:\n"
+        "  quit            Quit currently running app\n"
```
cgutman: ` Quit the currently running app`

app/cli/commandlineparser.cpp
```diff
@@ -179,13 +180,57 @@ GlobalCommandLineParser::ParseResult GlobalCommandLineParser::parse(const QStrin
...
+{
+}
+
+void QuitCommandLineParser::parse(const QStringList &args)
+{
+    CommandLineParser parser;
+    parser.setupCommonOptions();
+    parser.setApplicationDescription(
+        "\n"
+        "Quit currently running app at given host."
```
cgutman: `Quit the currently running app on the given host.`

app/gui/CliQuitStreamSegue.qml
```diff
@@ -0,0 +1,70 @@
...
+
+Item {
+    anchors.fill: parent
+
+    function onSearchingComputer() {
+        stageLabel.text = "Establishing connection to PC..."
+    }
+
+    function onQuitingApp() {
+        stageLabel.text = "Quiting app..."
```
cgutman: `Quitting` is misspelled here and in a few function names

app/main.cpp
```diff
@@ -369,5 +379,9 @@ int main(int argc, char *argv[])
 
     int err = app.exec();
 
+    // Give worker tasks time to properly exit. Fixes PendingQuitTask
+    // sometimes freezing and blocking process exit.
+    QThreadPool::globalInstance()->waitForDone(30000);
```
cgutman: This is interesting. Can you tell me more about what you were seeing here?

I think we do have some race conditions on process exit as I've seen a few random crashes locally deep inside Qt frameworks. I'm somewhat tempted to `exit(err)` at the end of this function to avoid the pointless crashes, but it would be nice to know what is actually going on at that time (since ComputerManager should stop polling in its destructor).

app/streaming/session.cpp
jhakonen: You mean that Session would expose the whole StreamingPreferences object? Not just this single option?

app/backend/computermanager.h
jhakonen: That would require for NvComputer to inherit from QObject, otherwise it cannot be called from QML. Should that be done instead?

app/backend/nvhttp.cpp
jhakonen: This at least happens when starting app from UI. The last time when NvComputer updates is at start of the stream where server state is  MJOLNIR_STATE_SERVER_LAUNCHING, which resets the app id to zero. When the session window is closed, quitApp() gets called just before NvComputer would update, which does nothing as app id is still zero.

app/cli/commandlineparser.cpp
jhakonen: Ok

app/cli/commandlineparser.cpp
jhakonen: Ok

app/gui/CliQuitStreamSegue.qml
jhakonen: Ok

app/main.cpp
jhakonen: I can fairly easily reproduce this with debug build without debugger. After ending the session moonlight.exe's process is left running even though its window is already closed. Attaching to the process shows these three threads which have Moonlight's sources visible in call stack:
main-thread
pending-quit-task
update-app-list
Unfortunately I don't have matching pdb files for Qt's binaries, so can't see where exactly the main thread is hanging at.

app/streaming/session.cpp
jhakonen: Note that I can't just use any StreamingPreferences object. I must be the one that is associated with the session as it may have values overwritten from command line.

app/gui/StreamSegue.qml
```diff
@@ -68,6 +91,12 @@ Item {
 
     onVisibleChanged: {
         if (visible) {
+            // Prevent session restart after execution returns from QuitSegue
+            if (sessionLaunched) {
```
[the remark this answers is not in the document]
jhakonen: We can, but it causes odd UI behaviour.

When starting an app from UI, and then closing the session, the UI briefly shows App View before jumping to PC View. The stack view has a brief transition animations where it flows from QuitSegue to AppView to PcView.

Without popping the StreamSegue the transitions flows from QuitSegue to StreamSegue and then to the PcView. The QuitSegue and StreamSegue are very similar to each other so the transition isn't as jarring.

cgutman:
Thanks. I'm going to push a few changes on top. Let me know if you find issues with them (here or on Discord).

PR #177, opened 2019-02-13
Fix hotkeys for non-latin keyboard layouts

konst2:


cgutman:
Can you explain how Moonlight's hotkeys currently behave on your non-latin keyboard and how you are expecting them to work?

konst2:
On my russian layout hotkeys don't works. No reaction on press Ctrl+Alt+Shift+X (for example) because Moonlight now receive Ctrl+Alt+Shift+Ч. If I want working hotkeys — I must start it only in «en» layout. It's very annoying.
I think — use scancodes here is better. They are layout-undependent.

cgutman:
Ok, if that's what other applications do. It sounds okay to me.

Thanks for the fix.

PR #181, opened 2019-02-18
SDL_HapticRumblePlay rumble for gamepad without SDL_HAPTIC_LEFTRIGHT support

konst2:
Some gamepads can rumble but haven’t SDL_HAPTIC_LEFTRIGHT support. My Xbox One controller connected to MacBook as example (I think, this is SDL bug).  Gamepads still can rumble, but only low-freq motor. It’s better than nothing! This pull request adds support for such gamepads  utilizing standard  SDL_HapticRumblePlay rumble method.


app/streaming/input.cpp
```diff
@@ -775,9 +775,20 @@ void SdlInputHandler::handleControllerDeviceEvent(SDL_ControllerDeviceEvent* eve
...
-        if (state->haptic != nullptr && (SDL_HapticQuery(state->haptic) & SDL_HAPTIC_LEFTRIGHT) == 0) {
-            SDL_HapticClose(state->haptic);
-            state->haptic = nullptr;
+        state->hapticMethod = GAMEPAD_HAPTIC_METHOD_NONE;
+        if (state->haptic != nullptr) {
+            if ((SDL_HapticQuery(state->haptic) & SDL_HAPTIC_LEFTRIGHT) == 0) {
+                if (SDL_HapticRumbleSupported(state->haptic)) {
+                    SDL_HapticRumbleInit(state->haptic);
+                    state->hapticMethod = GAMEPAD_HAPTIC_METHOD_SIMPLERUMBLE;
+                    SDL_HapticRumblePlay(state->haptic, 0.75, 350);
```
cgutman: Looks like `SDL_HapticRumblePlay()` was left here by mistake

app/streaming/input.cpp
```diff
@@ -775,9 +775,20 @@ void SdlInputHandler::handleControllerDeviceEvent(SDL_ControllerDeviceEvent* eve
...
         state->haptic = SDL_HapticOpenFromJoystick(SDL_GameControllerGetJoystick(state->controller));
         state->hapticEffectId = -1;
-        if (state->haptic != nullptr && (SDL_HapticQuery(state->haptic) & SDL_HAPTIC_LEFTRIGHT) == 0) {
-            SDL_HapticClose(state->haptic);
-            state->haptic = nullptr;
+        state->hapticMethod = GAMEPAD_HAPTIC_METHOD_NONE;
+        if (state->haptic != nullptr) {
+            if ((SDL_HapticQuery(state->haptic) & SDL_HAPTIC_LEFTRIGHT) == 0) {
+                if (SDL_HapticRumbleSupported(state->haptic)) {
+                    SDL_HapticRumbleInit(state->haptic);
```
cgutman: Only set `hapticMethod` if `SDL_HapticRumbleInit()` returns 0.

app/streaming/input.cpp
```diff
@@ -884,30 +895,42 @@ void SdlInputHandler::rumble(unsigned short controllerNumber, unsigned short low
...
+        // Play the new effect
+        m_GamepadState[controllerNumber].hapticEffectId = SDL_HapticNewEffect(haptic, &effect);
+        if (m_GamepadState[controllerNumber].hapticEffectId >= 0) {
+            SDL_HapticRunEffect(haptic, m_GamepadState[controllerNumber].hapticEffectId, 1);
+        }
+    } else if (m_GamepadState[controllerNumber].hapticMethod == GAMEPAD_HAPTIC_METHOD_SIMPLERUMBLE) {
+        SDL_HapticRumblePlay(haptic,
+                             std::min(1.0, std::max(highFreqMotor, lowFreqMotor)/65535.0),
+                             SDL_HAPTIC_INFINITY);
+        m_GamepadState[controllerNumber].hapticEffectId = 1;
```
cgutman: Setting `hapticEffectId` shouldn't be required for simple rumble, since we aren't using `SDL_HapticNewEffect()` so there's no allocated haptic effect to free later.

app/streaming/input.cpp
```diff
@@ -884,30 +895,42 @@ void SdlInputHandler::rumble(unsigned short controllerNumber, unsigned short low
...
-    if (m_GamepadState[controllerNumber].hapticEffectId >= 0) {
-        SDL_HapticRunEffect(haptic, m_GamepadState[controllerNumber].hapticEffectId, 1);
+        // Play the new effect
+        m_GamepadState[controllerNumber].hapticEffectId = SDL_HapticNewEffect(haptic, &effect);
+        if (m_GamepadState[controllerNumber].hapticEffectId >= 0) {
+            SDL_HapticRunEffect(haptic, m_GamepadState[controllerNumber].hapticEffectId, 1);
+        }
+    } else if (m_GamepadState[controllerNumber].hapticMethod == GAMEPAD_HAPTIC_METHOD_SIMPLERUMBLE) {
+        SDL_HapticRumblePlay(haptic,
+                             std::min(1.0, std::max(highFreqMotor, lowFreqMotor)/65535.0),
```
cgutman: I'm not sure that we want `std::max(highFreqMotor, lowFreqMotor)` here.

The low and high frequency motors have different sized weights on them, so they feel different. We should try to mimic that difference as well as we can given the API limitations.

On Android, where I have a similar issue, I scaled them so 75% of the amplitude was the low frequency motor and 25% was the high frequency motor. We should do something like that here (or if you have a better idea, I'm all ears).

app/streaming/input.cpp
```diff
@@ -775,9 +775,20 @@ void SdlInputHandler::handleControllerDeviceEvent(SDL_ControllerDeviceEvent* eve
         state->jsId = SDL_JoystickInstanceID(SDL_GameControllerGetJoystick(state->controller));
         state->haptic = SDL_HapticOpenFromJoystick(SDL_GameControllerGetJoystick(state->controller));
```
cgutman: It looks like you changed the file permissions on input.cpp from 644 to 755. Please put them back to 644.

app/streaming/input.cpp
konst2: I thought a small effect could be useful to understand which method was involved.

cgutman: I'd prefer a log message instead. It might get annoying for some users to have their gamepads rumble on their own every time they start a connected (especially if they have unused gamepads attached which may be resting on a table or other surface that could be loud when it rumbles).

konst2: Fixed

app/streaming/input.cpp
konst2: Fixed

app/streaming/input.cpp
konst2: Fixed

app/streaming/input.cpp
konst2: Fixed

app/streaming/input.cpp
konst2: I try make some tests with different weights — but need some time. Maybe there will be some ideas.

konst2: Fixed. I've tested some weights for motors and figured out, that hi freq motor is really lightweight. So  I've set 0.33 and 0.8 as weights (defines in input.h file). On my senses, it’s better then 0.25 and 0.75. But you can define that values on your own.

cgutman: Nice, thanks.

app/streaming/input.cpp
```diff
@@ -883,31 +895,42 @@ void SdlInputHandler::rumble(unsigned short controllerNumber, unsigned short low
...
-    if (m_GamepadState[controllerNumber].hapticEffectId >= 0) {
-        SDL_HapticDestroyEffect(haptic, m_GamepadState[controllerNumber].hapticEffectId);
-    }
+        if (m_GamepadState[controllerNumber].hapticMethod == GAMEPAD_HAPTIC_METHOD_LEFTRIGHT) {
+            if (m_GamepadState[controllerNumber].hapticEffectId >= 0) {
+                SDL_HapticDestroyEffect(haptic, m_GamepadState[controllerNumber].hapticEffectId);
+            }
+        } else if (m_GamepadState[controllerNumber].hapticMethod == GAMEPAD_HAPTIC_METHOD_SIMPLERUMBLE) {
+            SDL_HapticRumbleStop(haptic);
+        }
```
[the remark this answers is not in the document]
konst2: fixed. Thans for patience.

cgutman:
Thanks for the contribution!

PR #198, opened 2019-03-22
Fixes build for Steam Link SDK, works on AMD64

clee:
I can redo this to enable both alsa *and* pulseaudio on non-ARM systems, if desired, but there's no UI for selecting the audio backend so I'm not sure how important that is.

Anyway, this works for me building for both Steam Link and my local Debian amd64 box.

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -64,11 +64,10 @@ unix:!macx {
     PKGCONFIG += openssl sdl2 SDL2_ttf opus
 
     # For libsoundio
-    packagesExist(libpulse) {
-        PKGCONFIG += libpulse
-    }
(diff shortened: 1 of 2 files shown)


cgutman:
As we discussed on Discord, this is probably an issue with the Steam Link SDK itself. We also want libsoundio for ARM platforms running traditional distros like Raspbian, etc.

Anyway, I ended up fixing the issue by using SLAudio instead of libsoundio for Steam Link. https://github.com/moonlight-stream/moonlight-qt/commit/b7116657d9f2a733f81f38985f86335674401c3a

Thanks for your help on the Steam Link stuff.

Issue #211, opened 2019-04-09
Wayland support


xdbob:
**Is your feature request related to a problem? Please describe.**
Moonlight is almost the last piece of software keeping me to quit X11 completely

**Describe the solution you'd like**
A working accelerated renderer (I thought of VA-API) for wayland, particularly `sway`.

**Describe alternatives you've considered**
Staying on X11 a little while longer.

**Additional context**
I have started to work on it on my spare time. For now the the application crashes at the renderer launch but it seems to be a bug in QT (or sway) when hiding the Window (it can be worked around by removing `window.visible = false` in `StreamSegue.qml`).

Note: I use archlinux and sway as a compositor (and I set the adequate env vars to make SDL and QT use theirs respective Wayland backend)

Passed the crash the renderer window just never shows up.

Any comment, suggestion or help will be really appreciated.

d3m3vilurr:
more detail info;
before patch: process would be crashed.
after patch: case split
- if disabled VA-API(for example, set `DISPLAY` or didn't set `LIBVA_DRIVER_NAME` or `VDPAU_DRIVER`), randerer window work well.
- if enabled VA-API, renderer window never shows up.

cgutman:
I would like to understand why the crash occurs with Sway when the window is hidden before changing anything. I'm on Fedora 29 using Qt 5.11.3 and GNOME Shell 3.30.2 on Wayland, and I don't get any crash when starting the stream. The window hides properly when the stream begins.

As for the VAAPI crash, in my case it's because Mutter doesn't support YUV surfaces:
```
wl_surface@42: error 2: Failed to create a texture for surface 42: Unsupported buffer format 12760
request could not be marshaled: can't send file descriptorError sending request: Broken pipe
Aborted (core dumped)
```

The crash occurs deep inside vaGetSurfaceBufferWl().

d3m3vilurr:
I didn't setup the VA-API drvier at that time.
so, first patch fixed another crash. and might be only related sway compositor.
and I didn't get the crash after apply patch(+whether enabling VA-API or not).

cgutman:
After https://github.com/moonlight-stream/moonlight-qt/commit/b8b633a6ccfbf8771b034f526636caf12fb168b3 I have hardware acceleration with VAAPI on Wayland. We have to read the frame back to the CPU and use SDL to do the YUV-RGB conversion and render, but it's still way better than software decoding.

d3m3vilurr:
[...]

PR #212, opened 2019-04-11
Fix crash in wayland+sway environ

d3m3vilurr:
apply #211 solution

### DIFF
--- a/app/gui/StreamSegue.qml
+++ b/app/gui/StreamSegue.qml
@@ -32,7 +32,7 @@ Item {
         hintText.visible = false
 
         // Hide the window now that streaming has begun
-        window.visible = false
+        window.hideWindow()
     }
 
     function displayLaunchError(text)
@@ -59,7 +59,7 @@ Item {
         stackView.replace(stackView.currentItem, component.createObject(stackView, {"appName": appName}), StackView.Immediate)
 
         // Show the Qt window again to show quit segue
-        window.visible = true
+        window.showWindow()
     }
 
     function sessionFinished()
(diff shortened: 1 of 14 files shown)


xdbob:
Hum, it was (and is) a terrible and ugly hack. I don't think it's a good idea to apply it, at least without knowing the underlying reason.

app/streaming/session.cpp
```diff
@@ -982,6 +980,9 @@ void Session::exec(int displayOriginX, int displayOriginY)
         SDL_SetWindowFullscreen(m_Window, m_FullScreenFlag);
     }
 
+    // Pump the message loop to update the UI
+    emit connectionStarted();
```
d3m3vilurr: this is the main problem that application was crashed when starting streaming.
`window.visibility = false` would work when we apply only this part.
but exit stream still has problem.

app/gui/StreamSegue.qml
```diff
@@ -137,6 +137,9 @@ Item {
 
             // Run the streaming session to completion
             session.exec(Screen.virtualX, Screen.virtualY)
+
+            // Show the Qt window again after streaming
+            window.showWindow()
```
d3m3vilurr: in my test, i3ipc was never trigged when disconnecting time.
it would be related QT rendering & event part.
but I give up to find real reason, just forceful wake up the application.
(it could be breaking change)

d3m3vilurr:
this patch is still ugly hack. just prove the solution :)

xdbob:
Strange, it seems fixed on my computer. Probably by [qt5-wayland](https://code.qt.io/cgit/qt/qtwayland.git/tree/dist/changes-5.12.4/?h=v5.12.4)

PR #363, opened 2020-04-16
Fallback to ALSA on ARM64

PabloZaiden:
PulseAudio keeps crashing on ARM64 (tested on L4T Ubuntu, Jetson Nano and Gentoo).

This PR implements the same fallback used for the Raspberry PI

### DIFF
--- a/app/app.pro
+++ b/app/app.pro
@@ -190,6 +190,10 @@ HEADERS += \
     streaming/video/overlaymanager.h \
     backend/systemproperties.h
 
+unix:contains(QT_ARCH, arm64) {
+  DEFINES += ARM64
+}
+
 # Platform-specific renderers and decoders
 ffmpeg {
     message(FFmpeg decoder selected)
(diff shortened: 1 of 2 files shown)


PabloZaiden:
It has one downside (at least on L4T Ubuntu on the switch): volume cannot be changed with regular controls.
I'm assuming something could be done with alsamixer, but didn't try it yet.

Even with the downside, it's still better than having PulseAudio crashing every few minutes

PabloZaiden:
Still not sure if it’s better to have this by default on L4T or just make it able to enable it with an env var

cgutman:
I would prefer an environment variable since it seems like something the PulseAudio team is looking to fix soon and there are downsides like loss of volume control (and thanks for filing that PA bug by the way!).

Avoiding PA is doable with `SDL_AUDIODRIVER=alsa` today. Once the PA team fixes the bug, we can see if the Ubuntu team can backport the patch. If they won't, then we can have another discussion but for now I'd prefer to avoid this hack for now.

PabloZaiden:
Great. Didn’t know you could force ALSA right now. 
Let’s see what happens with the pulseaudio bug. They’ve added that as a blocker for 14.

PR #364, opened 2020-04-17
Fix use after free when a device controller is unloaded

AR-CADE:
This fix resolve an issue that resulting to a crash of the application when a device controller is unloaded (typically it should happen when a wired controller is unplugged or Bluetooth controller connection lost)

### DIFF
--- a/app/gui/sdlgamepadkeynavigation.cpp
+++ b/app/gui/sdlgamepadkeynavigation.cpp
@@ -1,4 +1,5 @@
 #include "sdlgamepadkeynavigation.h"
+#include <iostream>
 
 #include <QKeyEvent>
 #include <QGuiApplication>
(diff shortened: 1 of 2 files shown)


cgutman:
It's not clear to me where the use-after-free is occurring. Can you provide a stack trace for the crashing thread?

AR-CADE:
Yep:
[stack trace.txt](https://github.com/moonlight-stream/moonlight-qt/files/4531566/stack.trace.txt)

as you can see, it's crash in the onPollingTimerFired() callback method of sdlgamepadkeynavigation.cpp.

this because when unplugging a device controller, (at line 168) it use a SDL_GameController instance that has been already free'd by the SDL framework but still present in the m_Gamepads list.
By implementing the SDL_CONTROLLERDEVICEREMOVED case and by using a QMap instead a QList to store the SDL_GameController instance to store, retrieve or remove an instance, I managed to solve this issue :-)

cgutman:
I think that was a bug in SDL - https://bugzilla.libsdl.org/show_bug.cgi?id=5034

Can you try with the latest code from master which includes the fixed version of SDL?

AR-CADE:
thank you, I will look a it,

just give me few days to take some times to test it and report.

cgutman:
Any luck reproducing the issue with Moonlight v2.0.0 or master?

AR-CADE:
> Any luck reproducing the issue with Moonlight v2.0.0 or master?

Hi, sorry for the delay,
it's all good for me, update the SDL framework seem to do the job, I only noticed that I had to remap my gamepad for some games but it do not crash anymore 👍

PR #365, opened 2020-04-18
RFC: Introduce a new FFMPEG frontend renderer: EGLRenderer

### LINKED ISSUE #267: Add an OpenGL and/or Vulkan renderer
**Is your feature request related to a problem? Please describe.**
Currently on Wayland the rendering is done via SDL so the GPU computed frames are copied back to the main memory and then "magically" rendered. (see b8b633a6ccfbf8771b034f526636caf12fb168b3 and #211)

**Describe the solution you'd like**
Create a custom OpenGL or Vulkan based renderer that takes the VAAPI frames (in GPU mem) and does the color conversion directly within the GPU context. It should then be as efficient as the X11 version. (See how `mpv` does the rendering on wayland)

**Additional context**
I am willing to try and implement it but it's gonna take some time

xdbob:
Revision: 7 [!] Do not merge (unsigned commit)

Right now this renderer works on X11 & Wayland with VAAPI as a backend.

Some rendering latency benchmarks on my `i7-10510U` (with
`intel-media-driver` 20.1.1 which cause a *huge* regression with the
SDL_Renderer):
|                       | X11    | Wayland |
| Before            | 6.78ms | 22.50ms |
| EGLRenderer | 0.76ms | 00.77ms |

Closes #267 

PS: It doesn't support the overlay, but should be easy to implement

Changes
=======
* Since V6:

```
Antoine Damhet (7):
      eglvid: update TODO
      eglvid: fix assert condition
      eglvid: use linear scaling
      eglvid: fix typos
      eglvid: introduce macro to ease logging
      EGL: print more error codes on error
      eglvid: clear buffer before drawing
```

[full diff here](https://github.com/xdbob/moonlight-qt/compare/7676abb497690a04ffb5a2dcc9582f6ab4a2285d...34d61593f4631fb4ced9b478bd6ee6c82e94a1e5)

* since V5:

```
Antoine Damhet (1):
      egl: ffmpeg: fix my mess from b875a217c8958bf6d4102da62de45db44ca5a142
```

[full diff here](https://github.com/xdbob/moonlight-qt/compare/066363114a0ebac4651e837240292135558c8732...7676abb497690a04ffb5a2dcc9582f6ab4a2285d)

* since V4:

```
Antoine Damhet (8):
      EGL: misc: fix indent + eglvid destructor
      EGL: move `EGLExtensions` implementation into egl_extensions.cpp
      EGL: hack-up SDL OPENGL window via SDL_Renderer
      eglvid: fixup: m_dummyRenderer -> m_DummyRenderer
      eglvid: keep source video aspect ratio
      eglvid: set `SDL_HINT_OPENGL_ES_DRIVER` for X11
      eglvid: cosmetic: alphabetize includes
      EGL: Use the renderer only if direct rendering not supported (ie: wayland)
```

[full diff here](https://github.com/xdbob/moonlight-qt/compare/e6b39755288e2b6e0b1ff2e5c3a2177b3205cf2b...b875a217c8958bf6d4102da62de45db44ca5a142)

* since V3:

```
Antoine Damhet (7):
      EGL: cosmetics: intent + case fixes
      eglvid: load_and_build_shader -> EGLRenderer::loadAndBuildShader
      eglvid: explicit TODO for isRenderThreadSupported
      eglvid: avoid double-free for m_ShaderProgram
      eglvid: don't deinitialize in case of init failure
      eglvid: replace some fixups by asserts
      EGL: freeEGLImages now takes the image array as a parameter
```

[full diff here](https://github.com/xdbob/moonlight-qt/compare/4213856723ea3c59d2f9e0f876cc5ea7f4bdbde5...98ae79ba620f27b15f63aacb21bfc7964377e5eb)

* since V2:
```
Antoine Damhet (2):
      EGL: dynamically resole EGLImageTargetTexture2DOES to avoid link failures
      EGL/misc: mark unused argument by omitting their names
```

[full diff here](https://github.com/xdbob/moonlight-qt/compare/79bf494c0a7aad48d919bb5ebf0e16b7d1cbf0fa...650a12661c8c387f356362f2855d5696fa972a62)

* since V1:
```
Antoine Damhet (4):
      EGLRenderer: don't eglinit + deinit in case of init fail
      VAAPI: refacto of exportEGLImages failure
      EGLRenderer: remove renderOverlay & getDecoderColorspace
      EGLRenderer: cosmetic changes + indent fixup
```

[full diff here](https://github.com/xdbob/moonlight-qt/compare/b9cced7ec8a0592c5ce0ce1b415b9745ec509ba1...79bf494c0a7aad48d919bb5ebf0e16b7d1cbf0fa)

### DIFF
--- a/app/shaders/egl.frag
+++ b/app/shaders/egl.frag
@@ -1,0 +1,21 @@
+#version 300 es
+#extension GL_OES_EGL_image_external : require
+precision mediump float;
+out vec4 FragColor;
+
+in vec2 vTextCoord;
+
+uniform mat3 yuvmat;
+uniform vec3 offset;
+uniform samplerExternalOES plane1;
+uniform samplerExternalOES plane2;
+
+void main() {
+	vec3 YCbCr = vec3(
+		texture2D(plane1, vTextCoord)[0],
+		texture2D(plane2, vTextCoord).xy
+	);
+
+	YCbCr -= offset;
+	FragColor = vec4(clamp(yuvmat * YCbCr, 0.0, 1.0), 1.0f);
+}
--- a/app/streaming/video/ffmpeg-renderers/eglvid.cpp
+++ b/app/streaming/video/ffmpeg-renderers/eglvid.cpp
@@ -1,0 +1,480 @@
+// vim: noai:ts=4:sw=4:softtabstop=4:expandtab
+#define GL_GLEXT_PROTOTYPES
+
+#include "eglvid.h"
+
+#include "path.h"
+#include "streaming/session.h"
+#include "streaming/streamutils.h"
+
+#include <QDir>
+
+#include <Limelight.h>
+#include <unistd.h>
+
+#include <SDL_egl.h>
+#include <SDL_opengl.h>
+#include <SDL_opengles2.h>
+#include <SDL_opengles2_gl2ext.h>
+#include <SDL_render.h>
+#include <SDL_syswm.h>
+
+/* TODO:
+ *  - handle more pixel formats
+ *  - handle software decoding
+ */
+
+/* DOC/misc:
+ *  - https://kernel-recipes.org/en/2016/talks/video-and-colorspaces/
+ *  - http://www.brucelindbloom.com/
+ *  - https://learnopengl.com/Getting-started/Shaders
+ *  - https://github.com/stunpix/yuvit
+ *  - https://en.wikipedia.org/wiki/YCbCr#ITU-R_BT.601_conversion
+ *  - https://www.renesas.com/eu/en/www/doc/application-note/an9717.pdf
+ *  - https://www.xilinx.com/support/documentation/application_notes/xapp283.pdf
+ *  - https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.709-6-201506-I!!PDF-E.pdf
+ *  - https://www.khronos.org/registry/OpenGL/extensions/OES/OES_EGL_image_external.txt
+ *  - https://gist.github.com/rexguo/6696123
+ *  - https://wiki.libsdl.org/CategoryVideo
+ */
+
+#define EGL_LOG(Category, ...) SDL_Log ## Category(\
+        SDL_LOG_CATEGORY_APPLICATION, \
+        "EGLRenderer: " __VA_ARGS__)
+
+EGLRenderer::EGLRenderer(IFFmpegRenderer *backendRenderer)
+    :
+        m_SwPixelFormat(AV_PIX_FMT_NONE),
+        m_EGLDisplay(nullptr),
+        m_Textures{0},
+        m_ShaderProgram(0),
+        m_Context(0),
+        m_Window(nullptr),
+        m_Backend(backendRenderer),
+        m_VAO(0),
+        m_ColorSpace(AVCOL_SPC_NB),
+        m_ColorFull(false),
+        EGLImageTargetTexture2DOES(nullptr),
+        m_DummyRenderer(nullptr)
+{
+    SDL_assert(backendRenderer);
+    SDL_assert(backendRenderer->canExportEGL());
+}
+
+EGLRenderer::~EGLRenderer()
+{
+    if (m_Context) {
+        if (m_ShaderProgram)
+            glDeleteProgram(m_ShaderProgram);
+        if (m_VAO)
+            glDeleteVertexArrays(1, &m_VAO);
+        if (m_DummyRenderer)
+            SDL_DestroyRenderer(m_DummyRenderer);
+        SDL_GL_DeleteContext(m_Context);
+    }
+}
+
+bool EGLRenderer::prepareDecoderContext(AVCodecContext*, AVDictionary**)
+{
+    /* Nothing to do */
+
+    EGL_LOG(Info, "Using EGL renderer");
+
+    return true;
+}
+
+void EGLRenderer::notifyOverlayUpdated(Overlay::OverlayType)
+{
+    // TODO: FIXME
+}
+
+bool EGLRenderer::isRenderThreadSupported()
+{
+    // TODO: can we use DMA-BUF in multithreaded context ?
+    return false;
+}
+
+bool EGLRenderer::isPixelFormatSupported(int, AVPixelFormat pixelFormat)
+{
+    // Remember to keep this in sync with EGLRenderer::renderFrame()!
+    switch (pixelFormat)
+    {
+    case AV_PIX_FMT_NV12:
+        return true;
+    default:
+        return false;
+    }
+}
+
+int EGLRenderer::loadAndBuildShader(int shaderType,
+                                    const char *file) {
+    GLuint shader = glCreateShader(shaderType);
+    if (!shader || shader == GL_INVALID_ENUM) {
+        EGL_LOG(Error, "Can't create shader: %d", glGetError());
+        return 0;
+    }
+
+    auto sourceData = Path::readDataFile(file);
+    GLint len = sourceData.size();
+    const char *buf = sourceData.data();
+
+    glShaderSource(shader, 1, &buf, &len);
+    glCompileShader(shader);
+    GLint status;
+    glGetShaderiv(shader, GL_COMPILE_STATUS, &status);
+    if (!status) {
+        char shaderLog[512];
+        glGetShaderInfoLog(shader, sizeof (shaderLog), nullptr, shaderLog);
+        EGL_LOG(Error, "Cannot load shader \"%s\": %s", file, shaderLog);
+        return 0;
+    }
+
+    return shader;
+}
+
+bool EGLRenderer::compileShader() {
+    SDL_assert(!m_ShaderProgram);
+    SDL_assert(m_SwPixelFormat != AV_PIX_FMT_NONE);
+
+    // XXX: TODO: other formats
+    SDL_assert(m_SwPixelFormat == AV_PIX_FMT_NV12);
+
+    bool ret = false;
+
+    GLuint vertexShader = loadAndBuildShader(GL_VERTEX_SHADER, "egl.vert");
+    if (!vertexShader)
+        return false;
+
+    GLuint fragmentShader = loadAndBuildShader(GL_FRAGMENT_SHADER, "egl.frag");
+    if (!fragmentShader)
+        goto fragError;
+
+    m_ShaderProgram = glCreateProgram();
+    if (!m_ShaderProgram) {
+        EGL_LOG(Error, "Cannot create shader program");
+        goto progFailCreate;
+    }
+
+    glAttachShader(m_ShaderProgram, vertexShader);
+    glAttachShader(m_ShaderProgram, fragmentShader);
+    glLinkProgram(m_ShaderProgram);
+    int status;
+    glGetProgramiv(m_ShaderProgram, GL_LINK_STATUS, &status);
+    if (status) {
+        ret = true;
+    } else {
+        char shader_log[512];
+        glGetProgramInfoLog(m_ShaderProgram, sizeof (shader_log), nullptr, shader_log);
+        EGL_LOG(Error, "Cannot link shader program: %s", shader_log);
+        glDeleteProgram(m_ShaderProgram);
+        m_ShaderProgram = 0;
+    }
+
+progFailCreate:
+    glDeleteShader(fragmentShader);
+fragError:
+    glDeleteShader(vertexShader);
+    return ret;
+}
+
+bool EGLRenderer::initialize(PDECODER_PARAMETERS params)
+{
+    m_Window = params->window;
+
+    if (params->videoFormat == VIDEO_FORMAT_H265_MAIN10) {
+        // EGL doesn't support rendering YUV 10-bit textures yet
+        return false;
+    }
+
+    /*
+     * Create a dummy renderer in order to craft an accelerated SDL Window.
+     * Request opengl ES 3.0 context, otherwise it will SIGSEGV
+     * https://gitlab.freedesktop.org/mesa/mesa/issues/1011
+     */
+    SDL_SetHint(SDL_HINT_OPENGL_ES_DRIVER, "1");
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
+    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
+
+    int renderIndex;
+    int maxRenderers = SDL_GetNumRenderDrivers();
+    SDL_assert(maxRenderers >= 0);
+
+    SDL_RendererInfo renderInfo;
+    for (renderIndex = 0; renderIndex < maxRenderers; ++renderIndex) {
+        if (SDL_GetRenderDriverInfo(renderIndex, &renderInfo))
+            continue;
+        if (!strcmp(renderInfo.name, "opengles2")) {
+            SDL_assert(renderInfo.flags & SDL_RENDERER_ACCELERATED);
+            break;
+        }
+    }
+    if (renderIndex == maxRenderers) {
+        EGL_LOG(Error, "Could not find a suitable SDL_Renderer");
+        return false;
+    }
+
+    if (!(m_DummyRenderer = SDL_CreateRenderer(m_Window, renderIndex, SDL_RENDERER_ACCELERATED))) {
+        EGL_LOG(Error, "SDL_CreateRenderer() failed: %s", SDL_GetError());
+        return false;
+    }
+
+    SDL_SysWMinfo info;
+    SDL_VERSION(&info.version);
+    if (!SDL_GetWindowWMInfo(params->window, &info)) {
+        EGL_LOG(Error, "SDL_GetWindowWMInfo() failed: %s", SDL_GetError());
+        return false;
+    }
+    switch (info.subsystem) {
+    case SDL_SYSWM_WAYLAND:
+        m_EGLDisplay = eglGetPlatformDisplay(EGL_PLATFORM_WAYLAND_KHR,
+                                             info.info.wl.display, nullptr);
+        break;
+    case SDL_SYSWM_X11:
+        m_EGLDisplay = eglGetPlatformDisplay(EGL_PLATFORM_X11_KHR,
+                                             info.info.x11.display, nullptr);
+        break;
+    default:
+        EGL_LOG(Error, "not compatible with SYSWM");
+        return false;
+    }
+
+    if (!m_EGLDisplay) {
+        EGL_LOG(Error, "Cannot get EGL display: %d", eglGetError());
+        return false;
+    }
+
+    if (!(m_Context = SDL_GL_CreateContext(params->window))) {
+        EGL_LOG(Error, "Cannot create OpenGL context: %s", SDL_GetError());
+        return false;
+    }
+    if (SDL_GL_MakeCurrent(params->window, m_Context)) {
+        EGL_LOG(Error, "Cannot use created EGL context: %s", SDL_GetError());
+        return false;
+    }
+
+    const EGLExtensions egl_extensions(m_EGLDisplay);
+    if (!egl_extensions.isSupported("EGL_KHR_image_base") &&
+        !egl_extensions.isSupported("EGL_KHR_image")) {
+        EGL_LOG(Error, "KHR_image unsupported");
+        return false;
+    }
+
+    if (!m_Backend->initializeEGL(m_EGLDisplay, egl_extensions))
+        return false;
+
+    if (!(EGLImageTargetTexture2DOES = (EGLImageTargetTexture2DOES_t)eglGetProcAddress("glEGLImageTargetTexture2DOES"))) {
+        EGL_LOG(Error,
+                "EGL: cannot retrieve `EGLImageTargetTexture2DOES` address");
+        return false;
+    }
+
+    /* Compute the video region size in order to keep the aspect ratio of the
+     * video stream.
+     */
+    SDL_Rect src, dst;
+    src.x = src.y = dst.x = dst.y = 0;
+    src.w = params->width;
+    src.h = params->height;
+    SDL_GetWindowSize(m_Window, &dst.w, &dst.h);
+    StreamUtils::scaleSourceToDestinationSurface(&src, &dst);
+
+    glViewport(dst.x, dst.y, dst.w, dst.h);
+
+    glClearColor(0, 0, 0, 1);
+    glClear(GL_COLOR_BUFFER_BIT);
+
+    if (params->enableVsync) {
+        /* Try to use adaptive VSYNC */
+        if (SDL_GL_SetSwapInterval(-1))
+            SDL_GL_SetSwapInterval(1);
+    } else {
+        SDL_GL_SetSwapInterval(0);
+    }
+
+    SDL_GL_SwapWindow(params->window);
+
+    glGenTextures(EGL_MAX_PLANES, m_Textures);
+    for (size_t i = 0; i < EGL_MAX_PLANES; ++i) {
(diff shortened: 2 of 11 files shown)


app/backend/systemproperties.cpp
```diff
@@ -119,7 +119,13 @@ void SystemProperties::querySdlVideoInfo()
         }
     }
 
+#ifdef HAVE_EGL
+    SDL_Window* testWindow = SDL_CreateWindow("", 0, 0, 1280, 720, SDL_WINDOW_HIDDEN | SDL_WINDOW_OPENGL);
```
cgutman, line 123: I think you can get around having to modify all of these `SDL_CreateWindow()` calls with `SDL_WINDOW_OPENGL` by just creating a dummy SDL renderer in EGLRenderer using `SDL_CreateRenderer()` in EGLRenderer. SDL has a magic internal function called `SDL_RecreateWindow()` which can add `SDL_WINDOW_OPENGL` after the fact.

app/streaming/session.cpp
```diff
@@ -362,8 +369,24 @@ bool Session::initialize()
         return false;
     }
 
+#ifdef HAVE_EGL
+    /* Request opengl ES 3.0 context for EGL Renderer, otherwise it will
+     * SIGSEGV: https://gitlab.freedesktop.org/mesa/mesa/issues/1011
+     */
+    SDL_SetHint(SDL_HINT_OPENGL_ES_DRIVER, "1");
```
cgutman, line 376: Using `SDL_CreateRenderer()` you can explicitly specify your renderer driver without having to set a hint.

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+#include <SDL_opengl.h>
+#include <SDL_syswm.h>
+#include <SDL_opengles2.h>
+#include <SDL_opengles2_gl2ext.h>
+
+/* TODO:
+ *  - handle more pixel formats
+ *  - code refacto/cleanup
+ *  - handle software decoding
+ *  - handle window resize
```
cgutman, line 24: What currently happens if you resize?

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+{
+    deinitialize();
+}
+
+bool EGLRenderer::prepareDecoderContext(AVCodecContext*, AVDictionary**)
+{
+    /* Nothing to do */
+
+    SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION,
+                "Using EGL renderer!!");
```
cgutman, line 86: Remove the `!!`

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+void EGLRenderer::notifyOverlayUpdated(Overlay::OverlayType)
+{
+    // TODO: FIXME
+}
+
+bool EGLRenderer::isRenderThreadSupported()
+{
+    // TODO: FIXME (maybe ?)
+    return false;
```
cgutman, line 99: This determines whether rendering can be done on a separate thread (otherwise it is done on the main thread).

Does your renderer support that?

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+    SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION,
+                "Using EGL renderer!!");
+
+    return true;
+}
+
+void EGLRenderer::notifyOverlayUpdated(Overlay::OverlayType)
+{
+    // TODO: FIXME
```
cgutman, line 93: This is a minor regression from the current functionality of VAAPI on Wayland today, because the SDL renderer does have overlay support.

Do you think this is something you'd be able to finish up soon in a follow-up PR?

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+{}
+
+bool EGLExtensions::is_supported(const QString &extension) const {
+    return m_extensions.contains(extension);
+}
+
+EGLRenderer::EGLRenderer(IFFmpegRenderer *frontend_renderer)
+    :
+        m_SwPixelFormat(AV_PIX_FMT_NONE),
+        m_egl_display(nullptr),
```
cgutman, line 61: Member variables should use PascalCase like `m_FooBar` instead of underscores. Regular variables and function names should use `camelCase`.

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+    switch (pixelFormat)
+    {
+    case AV_PIX_FMT_NV12:
+        return true;
+    default:
+        return false;
+    }
+}
+
+static GLuint load_and_build_shader(GLenum shader_type,
```
cgutman, line 114: Please make this a class member function

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+                 file, shader_log);
+        return 0;
+    }
+
+    return shader;
+}
+
+bool EGLRenderer::compileShader() {
+    if (m_shader_program) {
+        glDeleteProgram(m_shader_program);
```
cgutman, line 142: Is there a code path where we can actually reach this case?

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+    glGetProgramiv(m_shader_program, GL_LINK_STATUS, &status);
+    if (status) {
+        ret = true;
+    } else {
+        char shader_log[512];
+        glGetProgramInfoLog(m_shader_program, sizeof (shader_log), nullptr, shader_log);
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION,
+                 "EGLRenderer: cannot link shader program: %s",
+                 shader_log);
+        glDeleteProgram(m_shader_program);
```
cgutman, line 180: Reset `m_shader_program` to 0 to avoid double-free in destructor.

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+    const EGLExtensions egl_extensions(m_egl_display);
+    if (!egl_extensions.is_supported("EGL_KHR_image_base") &&
+        !egl_extensions.is_supported("EGL_KHR_image")) {
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "EGL: KHR_image unsupported...");
+        deinitialize();
+        return false;
+    }
+
+    if (!m_frontend->initializeEGL(m_egl_display, egl_extensions)) {
+        deinitialize();
```
cgutman, line 248: These `deinitialize()` calls are redundant. The destructor will take care of cleaning up after failure.

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+                "EGLRenderer: unknown color space: %d, falling back to BT-601",
+                m_colorspace);
+    return bt601_lim;
+}
+
+bool EGLRenderer::specialize() {
+    if (!compileShader())
+        return false;
+    if (m_vao)
+        glDeleteVertexArrays(1, &m_vao);
```
cgutman, line 370: Can we have vertex arrays at this point?

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+    // XXX: Maybe we should keep the window ratio for the vertices
```
cgutman, line 372: What does this comment mean?

app/streaming/video/ffmpeg-renderers/renderer.h
```diff
@@ -68,4 +85,25 @@ class IFFmpegRenderer : public Overlay::IOverlayRenderer {
...
+    }
+
+    virtual ssize_t exportEGLImages(AVFrame *,
+                                    EGLDisplay,
+                                    EGLImage[EGL_MAX_PLANES]) {
+        return -1;
+    }
+
+    // Free the ressources allocated during the last `exportEGLImages` call
+    virtual void freeEGLImages(EGLDisplay) {}
```
cgutman, line 107: I think the caller should need to provide the `EGLImage` array. That way the backend renderer doesn't have to remember them itself.

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+EGLExtensions::EGLExtensions(EGLDisplay dpy) :
+    m_extensions(egl_get_extensions(dpy))
+{}
+
+bool EGLExtensions::is_supported(const QString &extension) const {
+    return m_extensions.contains(extension);
+}
+
+EGLRenderer::EGLRenderer(IFFmpegRenderer *frontend_renderer)
```
cgutman, line 58: The provided renderer pointer is actually the backend renderer that handles the decoding. EGLRenderer itself is the frontend renderer.

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+        return false;
+
+    GLuint fragment_shader = load_and_build_shader(GL_FRAGMENT_SHADER, "egl.frag");
+    if (!fragment_shader)
+        goto frag_error;
+
+    m_shader_program = glCreateProgram();
+    if (!m_shader_program) {
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION,
+                 "EGLRenderer: cannot create shader program");
```
cgutman, line 163: Indentation is off for some of these `SDL_Log` calls. These should be lined up with the `(`

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+        return QStringList();
+    }
+    return QString(egl_extensions_str).split(" ");
+}
+
+EGLExtensions::EGLExtensions(EGLDisplay dpy) :
+    m_extensions(egl_get_extensions(dpy))
+{}
+
+bool EGLExtensions::is_supported(const QString &extension) const {
```
cgutman, line 54: Can you move this `EGLExtensions` stuff into a separate file? It doesn't feel like it belongs here, since it's declared in `renderer.h`.

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,480 @@
...
+    if (SDL_GL_MakeCurrent(params->window, m_context)) {
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Cannot use created EGL context: %s",
+                     SDL_GetError());
+        return false;
+    }
+
+    const EGLExtensions egl_extensions(m_egl_display);
+    if (!egl_extensions.is_supported("EGL_KHR_image_base") &&
+        !egl_extensions.is_supported("EGL_KHR_image")) {
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "EGL: KHR_image unsupported...");
```
cgutman, line 242: Remove the `...`

app/streaming/video/ffmpeg-renderers/vaapi.cpp
```diff
@@ -442,3 +449,106 @@ VAAPIRenderer::renderFrame(AVFrame* frame)
...
+        EGLAttrib attribs[17] = {
+            EGL_LINUX_DRM_FOURCC_EXT, (EGLint)layer.drm_format,
+            EGL_WIDTH, i == 0 ? frame->width : frame->width / 2,
+            EGL_HEIGHT, i == 0 ? frame->height : frame->height / 2,
+            EGL_DMA_BUF_PLANE0_FD_EXT, object.fd,
+            EGL_DMA_BUF_PLANE0_OFFSET_EXT, (EGLint)layer.offset[0],
+            EGL_DMA_BUF_PLANE0_PITCH_EXT, (EGLint)layer.pitch[0],
+            EGL_NONE,
+        };
+        if (m_egl_ext_dmabuf) {
```
cgutman, line 513: What's the impact if `EGL_EXT_image_dma_buf_import_modifiers` is not supported?

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: There are two problems:
* the viewport cannot be shrinked (on sway at least, ie: the content exceed the window boundary if resizing down)
* I don't handle any kind of ratio so any window that would'nt match the src ratio would probably be stretched in some form

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: Sorry, old debug code

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: I kept this note because I am not sure of memory implications of the whole DMA-BUF management in a multithread context

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: To be honest I was kinda hoping someone else would do it :)
I don't really like OpenGL and displaying text seems a **lot** of boiler plate code.

My first approach of this feature (the wole PR) was to use the `SDLRenderer` and use `SDL_Texture` but the `YCbCr` planes cannot be directly bound.

Maybe we could hack around the `SDL_Renderer` to draw the text while doing raw GL before/after. If this works, it would take advantage of undocumented SDL behaviors and could break later, what do you think about it ?

PS: If needed I could do it using raw OpenGL, or better, if you agree I could use another new third party library to handle this (like [imgui](https://github.com/ocornut/imgui) for example)

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: Can the read-back pixel format change between two `renderFrame` call ?

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: `initialize` can only be called once (there will never be some retry on failure) ?

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: same question as https://github.com/moonlight-stream/moonlight-qt/pull/365#discussion_r420653168

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: It means that I should compute the ratio of the image I will display to keep the same ratio as the source, see https://github.com/moonlight-stream/moonlight-qt/pull/365#discussion_r420645039

app/streaming/video/ffmpeg-renderers/renderer.h
xdbob: You are right, I did this because the backend renderer is the owner of the frames, so in my mind it made sense that it would remember them

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: Yes, brainfart :)

app/streaming/video/ffmpeg-renderers/vaapi.cpp
xdbob: To be honest I am not sure (I don't have anything to test it) but  most DMA-BUF user I found have this fallback (wlroots, weston, kodi).

PS: the renderer still works on my computer if I disable this code

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,465 @@
...
+    if (!SDL_GetWindowWMInfo(params->window, &info)) {
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION,
+                     "!SDL_GetWindowWMInfo() failed: %s",
+                     SDL_GetError());
+        return false;
+    }
+    switch (info.subsystem) {
+    case SDL_SYSWM_WAYLAND:
+        m_EGLDisplay = eglGetPlatformDisplay(EGL_PLATFORM_WAYLAND_KHR,
+                                              info.info.wl.display, nullptr);
```
xdbob, line 220: I broke a few indentations while renaming variables (ie: FIXME)

app/shaders/egl.frag
```diff
@@ -0,0 +1,21 @@
+#version 300 es
+#extension GL_OES_EGL_image_external : require
```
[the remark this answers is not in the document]
xdbob: `egl.vert` is not specific to any format.

`egl.frag` (in its current state) should handle NV12 and NV21. In the long run when adding support for more pix fmts (YUV/YCbCr) I would like to use the same shader with `#ifdefs` in the `vec3 YCbCr` assignation.

xdbob:
Note: I think we should also update (or remove) the message "Detected Wayland. Performance may be worse than X11."

app/backend/systemproperties.cpp
xdbob: Is https://github.com/xdbob/moonlight-qt/commit/c940fedf244cfb992808c455c43125a49e2b39ed what you have in mind ?

cgutman: Yep, that cleans things up a lot. One minor nit: `m_dummyRenderer` should be `m_DummyRenderer`

cgutman:
> Note: I think we should also update (or remove) the message "Detected Wayland. Performance may be worse than X11."

Agreed.

app/streaming/video/ffmpeg.cpp
```diff
@@ -190,6 +194,15 @@ void FFmpegVideoDecoder::reset()
...
 bool FFmpegVideoDecoder::createFrontendRenderer(PDECODER_PARAMETERS params)
 {
+#ifdef HAVE_EGL
+    if (m_BackendRenderer->canExportEGL()) {
+        m_FrontendRenderer = new EGLRenderer(m_BackendRenderer);
+        if (m_FrontendRenderer->initialize(params)) {
+            return true;
+        }
+        delete m_FrontendRenderer;
+    }
```
cgutman, line 204: After thinking about this a bit, let's move this into the `!m_BackendRenderer->isDirectRenderingSupported()` case below. That will reduce the possible risk of regression to just the Wayland case which is already terrible today.

Once it has been in the wild for a bit, we can consider using EGL by default.

app/backend/systemproperties.cpp
xdbob: >< Sorry, I am not accustomed to letter case separations

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: As a mater of fact, yes (on `sway`, I will test it on `KWin` X11+Wayland soon enough).

sdl-resize2
sdl-resize

app/streaming/session.cpp
xdbob: Hum it seems that I need to keep the hint on X11, otherwise the `opengles2` renderer is not available (I can put it inside my `initialize()` method tho)

app/streaming/video/ffmpeg-renderers/eglvid.cpp
xdbob: * KWin(X11): resize works as expected
* KWin(Wayland): I just can't resize the window

cgutman: Ok, looks like a bug in SDL. Also not a regression going from SDLRenderer to EGLRenderer, so I'm okay with it.

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,495 @@
...
+    default:
+        return false;
+    }
+}
+
+int EGLRenderer::loadAndBuildShader(int shaderType,
+                                    const char *file) {
+    GLuint shader = glCreateShader(shaderType);
+    if (!shader || shader == GL_INVALID_ENUM)
+        return 0;
```
cgutman, line 112: Can you print an error here?

xdbob:
@cgutman 

> The code looks pretty good; performance is great. However, I'm seeing some strange artifacts sometimes when I resize the window where it will paint older window contents in the black letterbox/pillarbox areas. It doesn't always happen immediately after resizing (sometimes it can take a few minutes to see it) and it seems to flicker on and off (hard to catch with a screenshot).
> 
> Can you look into that?

Maybe I need to clear the output buffer before each frame, can you try ?
```
diff --git a/app/streaming/video/ffmpeg-renderers/eglvid.cpp b/app/streaming/video/ffmpeg-renderers/eglvid.cpp
index 3d4f308..1de2489 100644
--- a/app/streaming/video/ffmpeg-renderers/eglvid.cpp
+++ b/app/streaming/video/ffmpeg-renderers/eglvid.cpp
@@ -483,6 +483,7 @@ void EGLRenderer::renderFrame(AVFrame* frame)
         return;
     }
 
+    glClear(GL_COLOR_BUFFER_BIT);
     glUseProgram(m_ShaderProgram);
     glBindVertexArray(m_VAO);
     glDrawElements(GL_TRIANGLES, 6, GL_UNSIGNED_INT, 0);
@@ -490,4 +491,5 @@ void EGLRenderer::renderFrame(AVFrame* frame)
     SDL_GL_SwapWindow(m_Window);
     if (frame->hw_frames_ctx != nullptr)
         m_Backend->freeEGLImages(m_EGLDisplay, imgs);
+    glFinish();
 }
```

If it's not that I'm gonna dig into the SDL EGL implementation to see how exactly they handle their viewport

cgutman:
Yep, that was it. Just adding the `glClear()` was enough.

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,495 @@
...
+            EGLImageTargetTexture2DOES(GL_TEXTURE_EXTERNAL_OES, imgs[i]);
+        }
+    } else {
+        // TODO: load texture for SW decoding ?
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION,
+                     "EGL rendering only supports hw frames");
+        return;
+    }
+
+    glUseProgram(m_ShaderProgram);
```
cgutman, line 488: Need a `glClear(GL_COLOR_BUFFER_BIT)` here per our discussion

app/streaming/video/ffmpeg-renderers/eglvid.cpp
```diff
@@ -0,0 +1,495 @@
...
+
+    return shader;
+}
+
+bool EGLRenderer::compileShader() {
+    SDL_assert(!m_ShaderProgram);
+    SDL_assert(m_SwPixelFormat != AV_PIX_FMT_NONE);
+
+    // XXX: TODO: other formats
+    SDL_assert(m_SwPixelFormat != AV_PIX_FMT_NV12);
```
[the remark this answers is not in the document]
xdbob: Oh my, that means I haven't built in debug since.... a long time ago

xdbob:
In introduced an `EGL_LOG` macro (https://github.com/xdbob/moonlight-qt/commit/0848002a95141b3d05dce8e73d7ed0f4b94bcbb9) and functions to get human readable GL errors (https://github.com/xdbob/moonlight-qt/commit/1a94b19465bd38ecd496a1178c97cc5133321fae) are you ok with that @cgutman ?

cgutman:
Excellent work, merging per our discussion on Discord.

PR #433, opened 2020-08-23
Added option to switch primary and secondary mouse buttons

### LINKED ISSUE #325: Option to Reverse Primary/Second Mouse Clicks
I'm using a left-handed Razer Naga on my host computer and when logging into it remotely my mouse buttons are backwards.  When on my host computer I find that BIOS has the buttons reversed as well, so it's just an issue as a result of using a left handed mouse at home.  If we had some function to reverse the primary and secondary mouse buttons within Moolight this would fix this issue.

annk15:
Fixes #325

### DIFF
--- a/app/streaming/input/mouse.cpp
+++ b/app/streaming/input/mouse.cpp
@@ -55,6 +55,14 @@ void SdlInputHandler::handleMouseButtonEvent(SDL_MouseButtonEvent* event)
             return;
     }
 
+    if (m_SwapMouseButtons) {
+        if (button == BUTTON_RIGHT)
+            button = BUTTON_LEFT;
+        else if (button == BUTTON_LEFT)
+            button = BUTTON_RIGHT;
+    }
+
+
     LiSendMouseButtonEvent(event->state == SDL_PRESSED ?
                                BUTTON_ACTION_PRESS :
                                BUTTON_ACTION_RELEASE,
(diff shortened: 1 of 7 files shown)


cgutman:
Can you add a command-line option to enable this to CommandLineParser.cpp?

Please also remove your the `.DS_Store` you added.

annk15:
> Can you add a command-line option to enable this to CommandLineParser.cpp?
> 
> 
> 
> Please also remove the `app/.DS_Store` file you added.

Sure, I will fix that.

annk15:
I have added a command line option to toggle between the left/right button swap.
I also did some refactoring and removed the "app/.DS_Store" file 😃

app/streaming/input/mouse.cpp
```diff
@@ -55,6 +55,14 @@ void SdlInputHandler::handleMouseButtonEvent(SDL_MouseButtonEvent* event)
             return;
     }
 
+    if (m_SwapMouseButtons) {
+        if (button == SDL_BUTTON_RIGHT)
+            button = SDL_BUTTON_LEFT;
+        else if (button == SDL_BUTTON_LEFT)
+            button = SDL_BUTTON_RIGHT;
```
cgutman, line 62: These need to be using the Moonlight button definitions, not SDL's (see the switch statement above). This only works by luck (SDL's `SDL_BUTTON_LEFT` is the same value as `BUTTON_LEFT`)

annk15: I have now updated the code so it uses Moonlights button definitions.

cgutman:
Merged, thanks!

PR #451, opened 2020-09-30
French translation

watchha:


cgutman:
Merged, sorry for the delay

cgutman:
If you would like to update the translation, I have updated and added some strings.

PR #475, opened 2020-11-22
Update for French translation

watchha:


cgutman (changes requested):
I moved the languages into the `app` folder, please move your changes to that location.

It looks like you have some submodule changes in your branch too. Please remove those.

Issue #399, opened 2020-05-31
Start moonlight client in fullscreen mode


igno2k:
**Is your feature request related to a problem? Please describe.**
For a gaming pc (linux) hooked up to a tv I want to start the moonlight client in full screen mode without X11 windows decoration.

**Describe the solution you'd like**
Command line option to start moonlight client in fullscreen

jmt-gh:
I've opened PR #492 for this. I've added it as a UI option in the Moonlight UI, rather than a CLI option, as most of the options provided through the CLI are for the gamestream configuration, not the Moonlight UI

PR #492, opened 2020-12-29
Add support for Fullscreen UI in Moonlight

jmt-gh:
This PR adds support for the Moonlight UI to go in to decoration-less fullscreen (issue #399). ~~To try and remove confusion from the stream window mode, I went with `fullScreenModeUI` for the naming convention. As well, I unified the usage of Full-screen and Fullscreen with Fullscreen, as I think it reads a bit easier in a sentence.~~

To consolidate the UI, I implemented a Display Mode selector to allow users to choose between Windowed, Fullscreen (Windowed), and Fullscreen. 

Screenshot:
image
image

### DIFF
--- a/app/gui/SettingsView.qml
+++ b/app/gui/SettingsView.qml
@@ -721,19 +721,19 @@ Flickable {
                 }
 
                 CheckBox {
-                    id: kioskModeCheck
+                    id: fullScreenModeUICheck
                     width: parent.width
-                    text: qsTr("Start Moonlight in Kiosk Mode")
+                    text: qsTr("Start Moonlight in Full Screen Mode")
                     font.pointSize: 12
-                    checked: StreamingPreferences.kioskMode
+                    checked: StreamingPreferences.fullScreenModeUI
                     onCheckedChanged: {
-                        StreamingPreferences.kioskMode = checked
+                        StreamingPreferences.fullScreenModeUI = checked
                     }
 
                     ToolTip.delay: 1000
                     ToolTip.timeout: 5000
                     ToolTip.visible: hovered
-                    ToolTip.text: qsTr("Start Moonlight in full screen, with no window decorations.")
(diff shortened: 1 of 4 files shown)


jmt-gh:
Noting this is still a WIP -- trying to figure out a last bug, where moving from Fullscreen -> Fullscreen (windowed) puts you back in regular windowed mode instead

app/gui/SettingsView.qml
```diff
@@ -708,15 +708,49 @@ Flickable {
...
+                    id: uiDisplayModeComboBox
+                    textRole: "text"
+                    model: ListModel {
+                        id: uiDisplayModeListModel
+                        ListElement {
+                            text: qsTr("Windowed")
+                            val: StreamingPreferences.UI_WINDOWED
+                        }
+                        ListElement {
+                            text: qsTr("Fullscreen (Windowed)")
```
cgutman, line 743: Maybe we should just called it "Maximized"?

app/gui/SettingsView.qml
```diff
@@ -708,15 +708,49 @@ Flickable {
...
+                    text: qsTr("Display Mode")
                     font.pointSize: 12
-                    enabled: SystemProperties.hasWindowManager
-                    checked: !StreamingPreferences.startWindowed || !SystemProperties.hasWindowManager
-                    onCheckedChanged: {
-                        StreamingPreferences.startWindowed = !checked
+                    wrapMode: Text.Wrap
+                }
+
+                AutoResizingComboBox {
```
cgutman, line 719: It looks like this doesn't have the handling of having no window manager that the old code had

jmt-gh: Resolved this by setting a fallback in main.qml for when a window manager isn't present, and disabling this ability to change display mode here. Let me know if this doesn't catch the same corner case

app/gui/SettingsView.qml
jmt-gh: Done

PR #516, opened 2021-02-09
Add handlers for numpad ENTER (`Keys.onEnterPressed`)

maxiberta:
I tend to use the numpad's ENTER key to quickly confirm dialogs without having to lift, or even move, my right hand from the mouse. And it's a bit frustrating when it doesn't work, for example with Moonlight's quit dialog.

This is my admittedly naive attempt at fixing it: basically just duplicate all `Keys.onReturnPressed` handlers as `Keys.onEnterPressed` handler. It works locally (which is more that I expected!), but I'm not familiar with QML really, so will happily accept any suggestions.

Tested with a local snap build pointing at the proposed repo/branch on Ubuntu 20.10.

### DIFF
--- a/app/gui/AppView.qml
+++ b/app/gui/AppView.qml
@@ -245,6 +245,17 @@ CenteredGridView {
             }
         }
 
+        Keys.onEnterPressed: {
+            // Open the app context menu if activated via the gamepad or keyboard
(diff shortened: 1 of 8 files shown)


cgutman:
Yeah, I don't know if there's a cleaner way to do this either, but it seems fine for now.

PR #536, opened 2021-03-07
Add german translation

reloxx13:
Translated with QtLinguist-Standalone-5.12.0

Can someone make a test build for me?


cgutman:
It looks like the *.qm files are empty

reloxx13:
> It looks like the *.qm files are empty

no, github cannot display compiled/binary files.
grafik

SupervisedThinking:
You should squash your commits to have a cleaner pr 👍🏻

cgutman:
I'll squash on merge.

app/settings/streamingpreferences.h
```diff
@@ -63,6 +63,7 @@ class StreamingPreferences : public QObject
     enum Language
     {
         LANG_AUTO,
+        LANG_DE,
```
cgutman, line 66: This needs to go at the end of the list, otherwise it will cause existing language preferences to change due to renumbering of existing enum values.

cgutman (changes requested):

app/settings/streamingpreferences.h
cgutman: On second thought, I'll just fix this after merging and add a comment here so others don't make the same mistake.

Issue #448, opened 2020-09-26
Add commandline argument to list games on host


genova447:
**Is your feature request related to a problem? Please describe.**
It's not really a problem, just a thought I had that might be handy in some cases to have

**Describe the solution you'd like**
Add an argument called "list" to have Moonlight output a neat list of all games on the host.
So when using cmd for example to it, you might write:
*Moonlight.exe list hostname*

**Describe alternatives you've considered**
Otherwise maybe Moonlight could keep a log somewhere on the client PC of all the games found on that host (one log per host) since last time it was connected to it. It would just be a default function that it rewrites this log and keep it up to date, every time you manually connect to the host in question.

**Additional context**
I was simply thinking that it could be cool if Playnite (https://github.com/JosefNemec/Playnite) could have a plugin that can automatically add games to its library from whichever host you want. But I thought that, for that to happen, Moonlight would need some neat way to output a list/log of all the games currently found on the host.

Basically, in that case, Playnite would act as a frontend to Moonlight - a different way to start up games through Moonlight.

**Further explanation for other users**
In case it isn't super well-known, Moonlight has a cool feature to boot directly into games on a given host with the command:
*Moonlight.exe stream hostname/IP "gamename"* 
But currently for Playnite, you would have to manually add that string to the eligible entries in Playnite. But if this could be easily implemented, it would also make it easier to make a plugin for Playnite that can do all that automatically.
Though I am not myself able to make such plugin... so I would want to mention that maybe the feature would go entirely unused (don't want to waste developers' time on this if it's really not wanted of course).

But please excuse me however, if there already is an 'API' or some way for external programs to quickly find 'available games' via Moonlight... if there is, then disregard this and close the issue... I just didn't find that there would be a way to do it already.

**Bonus request**
Perhaps an ability to also 'search' for a game... If you happen to have multiple systems, and don't remember where a certain game is installed - you could (related to logging of games of each host) write:
*Moonlight.exe search "gamename"*
and it would search already known/approved hosts for all occurrences of the game.
I really don't know if anybody would have use of this... but I thought, if one feature is implemented, then maybe some people could find a use of this feature too.

PR #538, opened 2021-03-10
Add a 'list' option for the CLI

abusse:
This commit addresses Issue #448 by adding a command line option that allows the listing of all the Apps reported by the remote host as a CSV.

I wasn't sure what the best output format would be and what information to include. So for the first iteration I went will all available information about an App and the path to the cached box art as CSV. Since it is a simple printf that could be easily adapted if desired.


app/cli/listapps.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+                QCoreApplication::exit(0);
+            }
+            break;
+        }
+    }
+
+    void printApp(NvApp app) const
+    {
+        fprintf(stdout, "%s,%d,%s,%s,%s,%s,%s\n", qPrintable(app.name),
```
cgutman, line 117: This probably needs to handle escaping the app name, right?

app/cli/commandlineparser.cpp
```diff
@@ -464,3 +467,46 @@ QString StreamCommandLineParser::getAppName() const
...
+{
+    CommandLineParser parser;
+    parser.setupCommonOptions();
+    parser.setApplicationDescription(
+        "\n"
+        "List the available apps on the given host as CSV:\n"
+        "\tName, ID, HDR Support, App Collection Game, Hidden, Direct Launch, Path to Boxart"
+    );
+    parser.addPositionalArgument("list", "list available apps");
+    parser.addPositionalArgument("host", "Host computer name, UUID, or IP address", "<host>");
```
cgutman, line 489: CSV is nice for consuming in other applications but isn't very friendly for human consumption. I think `list` should just print the app names in a list by default and have a separate `--csv` flag for CSV output.

app/gui/CliListAppsSegue.qml
```diff
@@ -0,0 +1,57 @@
...
+import ComputerManager 1.0
+import Session 1.0
+
+Item {
+    function onSearchingComputer() {
+        stageLabel.text = qsTr("Establishing connection to PC...")
+    }
+
+    function onSearchingApps() {
+        stageLabel.text = qsTr("Searching for Apps...")
```
cgutman, line 13: Let's use `qsTr("Loading app list...")` here like CliStartStreamSegue does.

cgutman (changes requested):

app/cli/listapps.cpp
abusse: Currently, I cannot think about a situation that is not covered. Could you give an example of an input (app name) that you are concerned about to be escaped?

app/gui/CliListAppsSegue.qml
abusse: ACK

app/cli/listapps.cpp
cgutman: It's possible for app names to contain commas.

app/cli/listapps.cpp
```diff
@@ -102,25 +102,36 @@ class LauncherPrivate
...
+    void printAppCSV(NvApp app) const
     {
-        fprintf(stdout, "%s,%d,%s,%s,%s,%s,%s\n", qPrintable(app.name),
-                                               app.id,
-                                               app.hdrSupported ? "true" : "false",
-                                               app.isAppCollectorGame ? "true" : "false",
-                                               app.hidden ? "true" : "false",
-                                               app.directLaunch ? "true" : "false",
-                                               qPrintable(m_BoxArtManager->loadBoxArt(m_Computer, app).toDisplayString()));
+        fprintf(stdout, "\"%s\",%d,%s,%s,%s,%s,\"%s\"\n", qPrintable(app.name),
```
abusse, line 128: One corner case remains if the App name contains a `"` character. I already had code in place to escape that. However, when trying to test, I was not able to have an App with `"` in its name. Therefore, I assume that `"` is not possible in an App name. Furthermore, because I was not able to test the code, I decided to leave it out

cgutman:
I gave this a shot and there are some problems (not really the fault of your change, just how Moonlight currently works) that prevent it from being a good user experience right now IMHO.

1. It doesn't work at all on Windows. We are a Win32 subsystem app which means we don't have a console to write to unless we `AttachConsole()`. I'll need to look into that.
2. Even on platforms where it does work, like macOS and Linux, the output is full of log spam from polling PCs, initializing FFmpeg, etc. I'll need to add some way to maybe increase the minimum log level when running a command like `list` where terminal output is expected.
3. Because a Moonlight window appears when the `list` command is running, this pretty much precludes any background usage by launchers like Playnite. It's also just generally intrusive for human use too. I'm running `moonlight list` from the terminal to write output to that same terminal. Having windows flying around in the process of printing the console is really dumb (again, not your fault, just not something we've plumbed yet).

I'm going to skip this for v3.1.0 due to those issues, but I would like to get these addressed for the following release.

abusse:
I currently don't have a Windows system for testing to my disposal (nor have I done a lot Win32 programming in the past, tbh), so I cannot say much regarding that point. Regarding the second point, I only had this issues with a Debug build. With a Release build the output was "clean". Regarding the third point, that was I concern that I also had, however, since the `quit` command also opened a window, I assumed that was the way to go. I was concerned that a patch "circumventing" this behavior would be too intrusive and not be accepted for a merge. But since we seem to agree on that, I will have another look regarding that direction. Do you have already something in mind? If not, I would just start digging a little further when I have some more time to spare to come to a solution.

cgutman:
Ah yes true, on macOS and Windows release builds, we will redirect the normal error logging to a log file so it won't be printed to the console. I was testing on Linux which prints the output to the terminal.

We can probably use the same state machine and stuff, just printing progress messages to the console instead of using a window.

abusse:
I have made the changes, but I am not sure if that is what you had in mind. Let me know what you think.

cgutman:
Thanks, I'll take a look once I'm back from vacation

cgutman:
Rebased and merged manually to resolve conflicts:

de881769950bf34ed8af6ba8786dcfe378452f6d
0802609fd88341f37c34d197b0a6d569881a18cf
4ee36fd405f382315a494ac8464fb59aa5c44403

PR #551, opened 2021-03-21
Spelling: VSync

comradekingu:
As per https://en.wikipedia.org/wiki/Analog_television#Vertical_synchronization
Not quite sure though


cgutman:
I think V-Sync and VSync are both correct spellings for that term.

comradekingu:
@cgutman. "V-sync" makes sense in my head, but I don't really know. Seems there is no official resource.

cgutman:
I think I'll leave it the way it is for now.

Issue #213, opened 2019-04-12
Add profiles for fast settings switching


SpartanC90:
I would love too see the possibility to add different profiles.  The thing is, some games require a higher bitrate than other games to look good. Especially if you are on wireless with like a laptop, it would be nice to be able to switch settings on the fly instead of having to go to the settings and fiddle with the slider. 
Alternatively, it would be nice to be able to set up different settings for each game (if needed, they should pick global settings first).

ikogan:
In addition, some games would benefit from custom settings for just that game, similar to what https://github.com/moonlight-stream/moonlight-android/pull/919 is doing for Android.

One thing in particular is the ability to set "Remote Desktop" mode on a per-game basis. A game that's primarily controlled by the mouse could set that to avoid weird mouse acceleration issues, and an FPS could use the normal method.

tweelix:
For anyone that this may help. I currently have 2 separate "profiles" that I can choose by overwriting the $HOME environment variable on Linux. It will create new config files in this "fake" home directory and then choosing profiles is done using different shortcuts on the desktop that set up the environment.

ezamelczyk:
@tweelix how can I set that up on macOS? I can't find any config files and I literally did `find -name "*moonlight*" /`

Foul-Tarnished:
Also interested

Depending on the game, I would either run at 60fps or 120fps

Foul-Tarnished:
> @tweelix how can I set that up on macOS? I can't find any config files and I literally did `find -name "*moonlight*" /`

It doesn't work for me sadly

```
#!/bin/bash

export HOME=/Users/mia
Open -a "/Applications/Moonlight.app"
```

in a .sh file, set as executable (chmod +x filepath)
the settings are shared between this shortcut and the standard app
("mia" is another local account that exist, and isn't the one I'm using)

LetrixZ:
I just found out that this app has a CLI.

This is how I use it and does exactly what I need:

`/Applications/Moonlight.app/Contents/MacOS/Moonlight --1440 --fps 120 --bitrate 80000 --absolute-mouse --no-vsync --video-codec auto --video-decoder auto --display-mode borderless stream {HOST} "{APP}"`

Loyola-IS352:
[...]

PR #594, opened 2021-05-19
Add settings profiles

xbmo:
This commit implements a feature discussed in #213. 

It allows users to create settings profiles which store the changes they make on the settings menu under a user-defined name. Multiple profiles can be created, deleted and switched between. 

It has been implemented in a non-intrusive way for those who don't wish to use this feature, as the only thing that is different to their experience is the 'create new profile' box at the bottom of the settings menu.

create_profile

active_profile

profile_in_header

### DIFF
--- a/app/gui/SettingsView.qml
+++ b/app/gui/SettingsView.qml
@@ -54,6 +54,101 @@ Flickable {
         spacing: 15
 
         GroupBox {
+            id: profileChangeGroupBox
+            width: (parent.width - (parent.leftPadding + parent.rightPadding))
+            padding: 12
+            title: "<font color=\"skyblue\">" + qsTr("Change Profile") + "</font>"
+            font.pointSize: 12
+            visible: StreamingPreferences.hasProfiles
+            enabled: StreamingPreferences.hasProfiles
+
+            Column {
+                anchors.fill: parent
+                spacing: 5
+
+                Label {
+                    width: parent.width
+                    id: resActiveProfiletitle
+                    text: qsTr("Active Profile")
+                    font.pointSize: 12
+                    wrapMode: Text.Wrap
+                }
+
+                AutoResizingComboBox {
+                    function createModel() {
+                        var profilesListModel = Qt.createQmlObject('import QtQuick 2.0; ListModel {}', parent, '')
+
+                        var profilesList = StreamingPreferences.profiles
+                        
+                        for( var i in profilesList) 
+                        {
+                            profilesListModel.append({"text" : profilesList[i].name});
+                        }
+
+                        return profilesListModel
+                    }
+
+                    function reinitialize() {
+                        model = createModel()
+
+                        var activeProfileName = StreamingPreferences.activeProfileName
+                        currentIndex = 0
+                        for (var i = 0; i < model.count; i++) {
+                            var profileName = model.get(i).text
+                            console.log(profileName)
+
+                            // Pick the highest value lesser or equal to the saved FPS
+                            if (profileName == activeProfileName) {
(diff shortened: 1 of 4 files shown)


Proryanator:
I would love if this got merged in! Some games look amazing 4K@60, some games I prefer to play at 1080@120. This would be so helpful.

Proryanator:
In the mean time I'll probably pull your branch and start using it 😁

Proryanator:
Any traction on this PR? 😁

ZacTyAdams:
I was about to start work on this feature until I found this. Will this PR be completed anytime soon?

cgutman:
Yeah the main thing was basically the design of how profiles would work with different hosts and apps. It seems like we probably want a combination of per-app, per-host, and global settings with some sort of precedence system.

You probably want the freedom to make global changes to things like video decoder, finding PCs on the local network, UI language, etc. without having to go update every profile. For things like muting host PC speakers, you probably want that per-host. Settings like remote desktop mouse mode, HDR, and resolution probably tend to be per per-app.

However, there's a ton of overlap between per-app, per-host, and global settings. Bitrate might be a global or per-host setting, depending on if the network bottleneck is on the client or host side. Resolution and frame rate might be a global or per-app setting depending on your host PC performance and personal preference between image quality and performance. Reversing the scrolling direction might be global or per-host, depending on whether the host or client has inverted scrolling.

As a user, how do I create these profiles? How do I set precedence for each of these settings? How can I easily tell which combination of settings are the actual ones being applied (and if something isn't applied as expected, how can I figure out why?)

I don't have a clear idea answer to these design questions. If anyone has thoughts on this, I'd be happy to hear them.

ZacTyAdams:
I don't think you need to get that far into the weeds when it comes to the first iteration of the feature. A simple menu that offers the ability to snapshot your current host-level settings (all of them from host speakers, to bitrate, to resolution, etc) and give them a name is a perfect MVP. Most users are only connecting to one host but may have changes that need to be made depending on their client conditions like whether or not they are on wifi or ethernet, whether or not they need to change the resolution due to a particular app or game, or if they are enabling remote desktop mouse. I posted a message in Discord if there is any way I could help build this out as I am a user who definitely needs this with my client setup.

xbmo:
My thinking at the time was similar to what @ZacTyAdams has said, a MVP/simple approach to having this functionality available. 

A solution could go as deep as per-app, per-host etc, but from my experience as a user in emulators like Dolphin on Android, it comes with it's own issues such as it becoming difficult to keep track of the level at which something has been changed, and more per-app management. 

The way it's set up here, users aren't impacted by it unless it's a feature they want to interact with. At which point, there is some complexity in managing these profiles, but whatever solution is ultimately used will always have that to a certain degree.

ASchwad:
Any updates on this? Is there anything we can contribute to push this PR? 

> As a user, how do I create and manage these profiles? How do I set precedence for each of these settings? How can I easily tell which combination of settings are the actual ones being applied (and if something isn't applied as expected, how can I figure out why?)

As far as I understand the PR, profiles can be managed in the client settings. The applied settings are visible at the top before selecting a host. Personally, this is exactly what is was looking for - simple and straightforward.

Proryanator:
I'd also be interested; I often change res/fps/bitrates for different games.

Two-Play:
@xbmo @cgutman Hi, why hasn't this been merged yet? I just wanted to implement this feature and found this good PR. Can I do something to make it merged (faster)?

by the way, really cool project, thanks for all that

Proryanator:
+1 for getting this implemented (7 month check in here).

PR #598, opened 2021-05-21
Add Korean language

sanhoe:


cgutman:
Thanks, I integrated these changes in ebe9356937de957cd5057c24e9a466e31d5fd471 using the translation from Weblate.

If you have further changes, you can submit them through Weblate and they will be included automatically.

PR #676, opened 2021-11-18
enables H265 HDR option and adds PQ transferFunction 

shokakucarrier:
to render HDR stream on macOS 10.13+
Tested on macOS Monterey, it works.


cgutman:
Thanks!

PR #803, opened 2022-06-06
Use proper imperative in German

i-am-the-slime:


cgutman:
Thanks for the PR. We receive translations from [our Weblate project](https://hosted.weblate.org/projects/moonlight/moonlight-qt/). If you make the change there, it will get pulled in for the next release.

Issue #858, opened 2022-08-24
After upgrading SDL2 from 2.0.22-2 to 2.24.0-1, get crackling noise when running Moonlight


NebuPookins:
**Describe the bug**

I'm running Archlinux. I performed a system upgrade today, and there were no updates to Moonlight, but there was an update to the SDL2 library. After the update, whenever I connect to another machine via Moonlight, all audio produced by my machine becomes very crackly, including audio produced by other apps. For example, I can have a YouTube video running in the background, and as long as Moonlight isn't connected to another machine , the audio is fine. As soon as I connect to my remote desktop, the audio becomes crackly.

If I rollback the update to the SDL2 library, I am able to use Moonlight just fine with no audio distortion.

I'm not sure if this problem is specific to the way Moonlight is calling the SDL2 APIs. I was not able to reproduce the issue when using snes9x-gtk nor supertuxkart which I believe also use SDL2.

**Steps to reproduce**

- Have SDL 2.24.0-1 installed.
- Run moonlight.
- Connect to remote machine.

**Client PC details (please complete the following information)**
 - OS: Archlinux
 - Moonlight Version: 4.1.0-1 (but also occurs on older versions of Moonlight -- the issue is the SDL2 version, not the Moonlight version)
 - GPU: `[AMD/ATI] Ellesmere [Radeon RX 470/480/570/570X/580/580X/590] (rev e7)`

**Server PC details (please complete the following information)**
 - OS: Windows 10.0.19044 Build 19044
 - GeForce Experience version: 3.25.1.27
 - Nvidia GPU driver: 516.94

**Moonlight Logs (please attach)**

With the SDL2 2.0.24.0-1 (problem is present):

```
00:00:47 - SDL Info (0): Found host supported resolution: 1920x1080
00:00:47 - Qt Info: Executing request: "https://192.168.1.111:47984/resume?uniqueid=0123456789ABCDEF&uuid=7dc15b31618f432db88b11a6de10bb22&rikey=5d5e351f8468e1ee59bcc70e214afe8a&rikeyid=-1510896658&surroundAudioInfo=196610"
[...]

PR #859, opened 2022-08-25
Workaround SDL 2.24.0 issue with low sample rate

NebuPookins:
An audio issue was introduced when upgrading from SDL 2.0.22 to 2.24.0.
If we open an audio device in SDL with a low sample rate, this causes
all audio produced on the computer to become distorted (including audio
produced by apps other than Moonlight).

The workaround in this commit is simply to set a minimum sample rate of
480, at which point the problem disappears. The magic value 480 was
determined empirically, and slightly higher or slightly lower values
would probably work equally well. The default value before this
workaround was 240.

### DIFF
--- a/app/streaming/audio/renderers/sdlaud.cpp
+++ b/app/streaming/audio/renderers/sdlaud.cpp
@@ -1,5 +1,5 @@
 #include "sdl.h"
-
+#include <algorithm>
 #include <Limelight.h>
 #include <SDL.h>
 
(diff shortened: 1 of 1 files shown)


NebuPookins:
This is related to https://github.com/moonlight-stream/moonlight-qt/issues/858

Feel free to reject if you don't think this workaround is an appropriate fix.

cgutman:
Thanks for tracking this down. I checked in a very similar fix in aca82f400ac8cbc0ae98279a0b56751f9dd9d959 that just uses existing SDL macros and adds a comment explaining the issue.

PR #882, opened 2022-10-20
Update session.cpp

borewik:
This allows to use 5120 x 1440 common ultrawide resolution. 4k uses 8.8M pixels, Dual QHD around 7.4M.

### DIFF
--- a/app/streaming/session.cpp
+++ b/app/streaming/session.cpp
@@ -797,8 +797,8 @@ bool Session::validateLaunch(SDL_Window* testWindow)
         emitLaunchWarning(tr("An attached gamepad has no mapping and won't be usable. Visit the Moonlight help to resolve this."));
     }
 
-    // NVENC will fail to initialize when using dimensions over 4096 and H.264.
-    if (m_StreamConfig.width > 4096 || m_StreamConfig.height > 4096) {
+    // NVENC will fail to initialize when using dimensions over 8847360 pixels (4096 * 2160) and H.264.
+    if (m_StreamConfig.width*m_StreamConfig.height > 8847360) {
         // Pascal added support for 8K HEVC encoding support. Maxwell 2 could encode HEVC but only up to 4K.
         // We can't directly identify Pascal, but we can look for HEVC Main10 which was added in the same generation.
         if (m_Computer->maxLumaPixelsHEVC == 0 || !(m_Computer->serverCodecModeSupport & 0x200)) {


cgutman:
Can you explain more about what issue you're solving? 4096x4096 H.264 works fine for me when streaming from RTX 2080.

borewik:
If you use ultrawide monitor like CRG9 or Neo G9, the native resolution is 5120 x 1440. 5120 is more than 4096 so it throws error despite the fact that eg. my 2080 TI can easily handle 5120 x 1440 since it requires less bandwidth than full 4k. This fixes that problem and might help few others like me.

cgutman:
It shouldn't throw an error if your client and host supports HEVC (which should be the case for the 2080 Ti). Which error are you getting?

borewik:
I was getting "Your host PC's GPU doesn't support streaming video resolutions over 4K." error.
I'm using Sunshine on Win10 as a host. "Found encoder nvenc: [h264_nvenc, hevc_nvenc]" shows up in ss logs so I assume it works. On the guest side I'm running Fedora, linux Moonlight client.
But correct me if I'm wrong - checking width > 4096 || height > 4096 is faulty  beacuse you could use some custom resolution like 4090x4090 that would exceed 4K bandwidth (amount of pixels) almost twice, and pass this check.

cgutman:
> I was getting "Your host PC's GPU doesn't support streaming video resolutions over 4K."

Ah, the issue is that Sunshine is not advertising support for the HEVC Main 10 codec which Moonlight uses to identify the NVENC generation.

It looks like Sunshine's logic for auto-selecting HEVC mode doesn't take into account reporting support for HEVC Main 10. That's not too surprising because I don't think Sunshine supports HDR, but it does have this unintended consequence of breaking Moonlight's logic for checking 8K streaming support.

If you set the HEVC mode to 3 in the Sunshine settings, it should start working.

> But correct me if I'm wrong - checking width > 4096 || height > 4096 is faulty beacuse you could use some custom resolution like 4090x4090 that would exceed 4K bandwidth (amount of pixels) almost twice, and pass this check.

The check itself is correct but the message is wrong. It means 4Kx4K where it just says 4K.

When commenting out the check and forcing H.264:
4096x4096 - works
4098x4098 - fails
4098x4096 - fails
4096x4098 - fails
1000x4098 - fails
4098x1000 - fails

cgutman:
I modified the resolution check to be ignored on unofficial hosts like Sunshine.

PR #893, opened 2022-11-07
Fix translation bug

happyharryh:
I found two imperfections related to the translation.

- First:

  The translation of the standard buttons in Dialogs (`Yes`, `No`, `OK`, `Help`, `Cancel`) is lacking for some languages.

  

  It is a bug (or temporary incompletion) of Qt: `windeployqt.exe` does not create `%BUILD_ROOT%\translations\qt_*.qm` for some languages due to the lack of `%QT_ROOT%\%QT_VERSION%\msvc2019_64\translations\qtbase_*.qm` ([Bug Report](https://bugreports.qt.io/browse/QTBUG-85840)). For Qt 5.15.2, the involved languages are:
  ```
  zh_CN, nb_NO, vi, th, nl, sv, pt, pt_BR, el, hi
  ```
  For the newest Qt 6.4.0, `zh_CN`, `nl` and `pt_BR` have been added, while the others are still lacking. A tricky way to fix it is to add translation for the languages into our `app/languages/qml_*.qm`, like the commit in this pull request. Another way is to create and deploy `%BUILD_ROOT%\translations\qt_*.qm` for the languages by ourselves. Actually, if `%BUILD_ROOT%\translations\qt_*.qm` exists, Qt will use an inner translator of the system language for the buttons. Even if we select another language in the settings view (reinstall the translator in `StreamingPreferences`), the buttons are always using the system language, which may be strange. So should we consider not deploying `%BUILD_ROOT%\translations\qt_*.qm` and using `app/languages/qml_*.qm` for all the languages?

- Second:

  `Language::LANG_AUTO` does not work for Simplified Chinese. It is because the return of `QLocale::system().name()` (`zh_CN`) is not strictly equal to the language suffix used in moonlight-qt (`zh_cn`), and `QTranslator::load` is case-sensitive. I have fixed it by changing `zh_cn` to `zh_CN`. By the way, I am not sure if there is a same problem with Traditional Chinese (`zh_TW` != `zh_Hant`?).


cgutman:
I think you're right about Traditional Chinese. I will change `zw_Hant` to `zw_TW`.

PR #957, opened 2023-02-15
Update bug and feature_request forms

bladeoner:
@cgutman this PR replaces the md files for yaml files and creates an easier way to submit a bug or feature.

It will automatically add the bug or enhancement label, which makes it easier to sort and/or filter.


bladeoner:
@cgutman I have just updated the PR, the reports are now tailored for a client pc.

bladeoner:
@cgutman when you have some time can you take a look at the PR?

Thanks in advance.

Issue #939, opened 2023-01-25
[Bug] Moonlight process stays running on stream termination


Chris3773:
**Describe the bug**
When using Sunshine as the host and launching an app from the command line `Moonlight.exe` will stay running when the stream terminates. The window is closed but the `Moonlight.exe` process is still running. When a stream is terminated, the app exited from within the stream, Sunshine handles the termination differently then GeForce Experience. Maybe a problem on both sides as GeForce Experience does not have this problem. In any case the `Moonlight.exe` process should terminate when this happens.

**Steps to reproduce**

1. Run the Sunshine host.
2. Launch an app from the cmd: Moonlight.exe stream "{Computer}" "{App}"
3. Close the app within the stream.
4. Check that the 'Moonlight.exe' is still running on the system.

**Other Moonlight clients**
The Android app will show a terminated alert when the app is terminated within the stream.

**Moonlight settings (please complete the following information)**
This does happen with default settings.

**Client PC details**
 - OS: Windows 10 19045
 - Moonlight Version: 4.3.1
 - GPU: Intel Iris Xe Graphics

**Server PC details**
 - OS: Windows 10 19045
 - Sunshine version: v0.17.0
 - Nvidia GPU driver: 528.24
 - Antivirus and firewall software: Windows Defender and Windows Firewall

**Moonlight Logs**
[moonlight.log](https://github.com/moonlight-stream/moonlight-qt/files/10498729/moonlight.log)

bocki:
I'm not completely sure this is unique to sunshine as this is happening to me as well, but using GFE and using the latest Linux CI AppImage (for testing purposes), last line of the log is exactly the same:

```
00:00:09 - SDL Info (0): Dropping window event during flush: 6 (1280 800)
00:00:09 - SDL Info (0): IDR frame request sent
00:00:09 - FFmpeg: [hevc @ 0x7f62fc658800] Output frame with POC 0.
00:00:09 - SDL Info (0): VDPAU surface size: 1280x800
00:00:52 - SDL Info (0): Control stream received unexpected disconnect event
00:00:52 - SDL Error (0): Connection terminated: -1
00:00:52 - SDL Info (0): Quit event received
00:00:52 - SDL Info (0): Failed to send ENet control packet
00:00:52 - SDL Info (0): Loss Stats: Transaction failed: 0
00:00:52 - SDL Info (0): Global video stats
00:00:52 - SDL Info (0): ----------------------------------------------------------
[...]

PR #977, opened 2023-03-31
Do not show error dialog if quitAfter is set

Chris3773:
If quitAfter is set for the session it will quit without trying to display the error dialog. The window was not visible when the dialog was displayed causing the Moonlight to stay running after the session ended.

The window could be set to be visible to display the error dialog however Sunshine returns an unexpected termination error when the app that the session is bound to is closed. Which would give you the error dialog every time you close the app to end the session. This is why I went with removing the error dialog instead of displaying it as the use case for this is launching directly into a session from the command line without the GUI. 

Fix: #939

### DIFF
--- a/app/gui/StreamSegue.qml
+++ b/app/gui/StreamSegue.qml
@@ -78,15 +78,8 @@ Item {
         SdlGamepadKeyNavigation.enable()
 
         if (quitAfter) {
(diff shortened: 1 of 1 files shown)


cgutman:
I believe the underlying issue causing lack of termination after dismissal of the error dialog has been fixed.

PR #1015, opened 2023-05-27
Add controller motion support

FallingSnow:
I wanted to add support for client side cursors (https://github.com/LizardByte/Sunshine/discussions/1276) but figured that might be a little complicated. So I decided to start with another issue I was having, controller motion. I play Zelda: BOW on CEMU via moonlight. There are areas that require controller motion input which I can test on.

I know that this will probably require some tweaks to sunshine too. I had a couple of questions/comments:

- SDL_GameControllerGetSensorDataRate reports 1000hz for the PS5 controllers gyro. 1000pps might be a lot so can I reduce this to say 2x target frame rate or maybe just 60pps?
- It does appear that ViGEmBus supports motion: https://github.com/ViGEm/ViGEmBus/issues/11
- What kind of sunshine changes would need to be made?
- Would I just add gyro info to [NV_MULTI_CONTROLLER_PACKET](https://github.com/moonlight-stream/moonlight-common-c/blob/master/src/Input.h#L87) to send it over the network? What's the best way to do this while maintaining backwards compatibility?

### DIFF
--- a/app/streaming/input/gamepad.cpp
+++ b/app/streaming/input/gamepad.cpp
@@ -96,6 +96,17 @@ Uint32 SdlInputHandler::mouseEmulationTimerCallback(Uint32 interval, void *param
     return interval;
 }
 
+void SdlInputHandler::handleControllerSensorEvent(SDL_ControllerSensorEvent* event)
+{
+    // Currently unhandled
+    switch (event->sensor)
+    {
+        case SDL_SENSOR_ACCEL:
+            SDL_LogVerbose(SDL_LOG_CATEGORY_APPLICATION,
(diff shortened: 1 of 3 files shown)


cgutman:
Thanks, I ended up including a controller motion packet extension in my latest batch of Sunshine protocol extension in moonlight-common-c.

https://github.com/moonlight-stream/moonlight-common-c/blob/812ec0e2b7d1a84c13ebae5f821aeedc80bd1ea2/src/Limelight.h#L461-L466
https://github.com/moonlight-stream/moonlight-common-c/blob/812ec0e2b7d1a84c13ebae5f821aeedc80bd1ea2/src/Limelight.h#L704-L710

PR #1063, opened 2023-09-03
Allow to specify gamepad GUIDs to be ignored when streaming

FrogTheFrog:
I want to be able to exclude certain GUID from being used as an input so I made this PR :). Also improved the log output a little to provide more info about the used device.

### DIFF
--- a/app/streaming/input/input.h
+++ b/app/streaming/input/input.h
@@ -179,6 +179,7 @@ private:
     bool m_FakeCaptureActive;
     QString m_OldIgnoreDevices;
     QString m_OldIgnoreDevicesExcept;
+    QStringList m_IgnoreDeviceGuids;
     StreamingPreferences::CaptureSysKeysMode m_CaptureSystemKeysMode;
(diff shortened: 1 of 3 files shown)


app/streaming/input/input.h
```diff
@@ -179,6 +179,7 @@ class SdlInputHandler
     bool m_FakeCaptureActive;
     QString m_OldIgnoreDevices;
     QString m_OldIgnoreDevicesExcept;
+    QStringList m_ignoreDeviceGuids;
```
cgutman, line 182: `m_IgnoreDeviceGuids` to match capitalization of other fields there.

cgutman (changes requested):
You should also add a check for `m_IgnoreDeviceGuids` in `SdlInputHandler::getAttachedGamepadMask()`, otherwise we'll still report that gamepad to the host when we initially connect.

FrogTheFrog:
Done

cgutman (approved):

PR #1077, opened 2023-10-02
Experimental beginnings of adding Hotkeys to eventually allow jumping back and forth between apps

paulpv:
This is just a conversation piece right now.

The short-term goal is to have Moonlight register global hotkeys for apps.
The proposed workflow is:
1. Browse the PCs and apps like normal
2. Right-Click on a PC's app and `Add Hotkey`
3. Use `Ctrl-Alt-Shift-H` to go view the Hotkeys
4. Eventually add ability to manage (rearrange, delete, etc) Hotkeys
5. Use `Control-Alt-Shift-X` (X = `1` to `0`) to launch the pre-defined PC's app.

This current code does NOT:
1. Manage (rearrange, delete hotkeys) in the Hotkeys view
2. Hotkeys are not yet "Global" and only have affect in the Hotkeys view
3. Hotkeys don't actually launch anything yet (they are just a placeholder for future launching)

This code is admittedly [hopefully only] a little sloppy, and is mostly just for conversation.


paulpv:
In English_US, the words "Computer(s)" and "PC(s)" seemed to be used interchangeably.

I noticed in other languages that sometimes they are translated as "Hosts", which actually makes a little more sense (since the host could possibly not be a Computer or PC).

I had already settled on "Ctrl+Alt+Shift+H" as a shortcut for the "Hotkeys" view, so I also chose "Ctrl+Alt+Shift+P" for the PCs view.
If "PCs" were to be renamed to "Hosts" then "Ctrl+Alt+Shift+P" would not make sense, and "Ctrl+Alt+Shift+H" would not make sense for Hotkeys.

So, perhaps a better choice would be:
* `Ctrl+Alt+Shift+H` for Hosts
* `Ctrl+Alt+Shift+K` for Hotkeys

paulpv:
Renamed this branch from the overly ambitious `agent-smith` to the less ambitious `hotkeys`.
This invalidates the PR, so I will open a new draft one soon.

PR #1105, opened 2023-11-01
Update moonlight-common-c to eb67a2fd (2023-10-26)

cathyjf:
Ever since I upgraded to moonlight-qt version 5.0.0, the Moonlight client has been unusable on my computer. Specifically, although I can connect to the host successfully, the connection inevitably and inexplicably terminates after a matter of seconds. After playing around `git bisect`, I determined that the problem had something to do with changes to how Moonlight handles host computers known by multiple IP addresses (with a different routing table for each IP).

I then discovered that you had already identified and fixed this problem in your recent commits to moonlight-stream/moonlight-common-c. I can confirm that the latest version of moonlight-stream/moonlight-common-c solves the problems and restores proper functionality to the moonlight-qt client for me.

This pull request updates the `moonlight-common-c` submodule of `moonlight-qt` to point to the latest version of moonlight-stream/moonlight-common-c. Because I can and have built my own version of moonlight-qt, I am not personally inconvenienced by this bug any longer, but it's a pretty serious bug insofar as it completely breaks the Moonlight client for people with certain networking configurations. Therefore, I suggest that you may want to release a new version of moonlight-qt based on this updated code (perhaps version 5.0.1?).

I can also confirm that the same problem exists in moonlight-stream/moonlight-ios@HEAD. That repository should also update its version of `moonlight-common-c`; however, that problem is less urgent because you haven't pushed a new version of `moonlight-ios` to the App Store in a while anyway.


cathyjf:
cc @cgutman

cgutman:
Thanks for the report and PR. I do intend to release v5.0.1 soon with this fix.

PR #1112, opened 2023-11-08
Add option to ignore input from Guide button

FrogTheFrog:
This option allows to disable the 'Guide' button passthrough to the host.

It is very useful when using SteamDeck with other controllers as pressing the Guide button activates stuff on both SteamDeck (in gamemode only when steam input is disabled, in desktop mode - always :/).

Guide button on the host can be emulated by Sunshine or other means.

Note: wanted to use "disable" instead of "ignore", but then people might misunderstand that it will also be disabled on the client somehow.


FrogTheFrog:
Need a better solution as the normal input is still being passed...

PR #1139, opened 2023-12-14
Add option to enforce a minimum latency to improve frame pacing

kirksaunders:
## Problem
If a user has a poor network connection (high jitter), they may experience instability in frame delivery. This manifests as stutters, and overall is not a great experience. This problem is exacerbated when frames are arriving near the end of the client's vsync period (since there's less time to present the frame before we roll over to the next vsync period).

## Solution
To account for jitter in the network connection (or in the host PC capture rate), we can allow the user to purposefully buffer some frames. This trades latency for stability. However, we need to be careful with how the buffering is done. If we simply render a frame once the next is available, that gives us one frame of buffer. However, that doesn't solve the frame pacing issue at all, since the second frame's arrival time dictates the render time. We can't rely on any individual frame's arrival time for timing our frame rendering.

Instead, we need to have a consistent schedule for releasing frames for rendering. We can measure the average time a frame spends in the input queue before it is released for rendering. Based on that measurement, we can gradually adjust our frame release schedule to target a desired minimum latency. Benefits of this approach are:
- The frame release schedule should be consistent, and mostly independent of individual frame delivery timing.
- The amount to buffer is granular. You could buffer 2 milliseconds, 6, milliseconds, 50 milliseconds, etc. It doesn't need to be a multiple of the frame period.
- With the gradual adjustment of the release schedule, it adapts to changes in the frame delivery timing/incoming framerate.

This "minimum latency" approach is exactly what is implemented in this PR. (I'm also open to different naming... Maybe "buffered latency"?)

## Related Items
Idea on Moonlight Ideas board: https://ideas.moonlight-stream.org/posts/251/frame-buffering-option

## Testing
I've tested this (and other methods for smoothing frame delivery) extensively, and found this gives me the best experience. My hardware is:
- Powerful Host PC (Windows 11, RTX 3070, 5950x)
- Steam Deck as client (moonlight-qt flatpak)
- Wired gigabit from host to router
- Wifi 6 from client to router

I typically use this option with 6 milliseconds of minimum latency, vsync enabled, and frame pacing disabled (not supported on Steam Deck anyways).

 I'm looking for others to test with various hardware configurations to see whether it offers a net improvement.

### DIFF
--- a/app/cli/commandlineparser.cpp
+++ b/app/cli/commandlineparser.cpp
@@ -374,6 +374,7 @@ void StreamCommandLineParser::parse(const QStringList &args, StreamingPreference
     parser.addChoiceOption("capture-system-keys", "capture system key combos", m_CaptureSysKeysModeMap.keys());
     parser.addChoiceOption("video-codec", "video codec", m_VideoCodecMap.keys());
     parser.addChoiceOption("video-decoder", "video decoder", m_VideoDecoderMap.keys());
+    parser.addValueOption("minimum-latency", "Minimum latency");
 
     if (!parser.parse(args)) {
         parser.showError(parser.errorText());
@@ -498,6 +499,13 @@ void StreamCommandLineParser::parse(const QStringList &args, StreamingPreference
     // Resolve --video-decoder option
     if (parser.isSet("video-decoder")) {
         preferences->videoDecoderSelection = mapValue(m_VideoDecoderMap, parser.getChoiceOptionValue("video-decoder"));
+    }
+
+    if (parser.isSet("minimum-latency")) {
+        preferences->minimumLatency = parser.getIntOption("minimum-latency");
+        if (!inRange(preferences->minimumLatency, 0, 50)) {
+            parser.showError("Minimum latency must be in range: 0 - 50");
+        }
     }
 
     // This method will not return and terminates the process if --version or
--- a/app/gui/SettingsView.qml
+++ b/app/gui/SettingsView.qml
@@ -812,6 +812,45 @@ Flickable {
                     ToolTip.visible: hovered
                     ToolTip.text: qsTr("Frame pacing reduces micro-stutter by delaying frames that come in too early")
                 }
+
+                Label {
+                    width: parent.width
+                    id: minimumLatencyTitle
+                    text: qsTr("Minimum latency:")
+                    font.pointSize: 12
+                    wrapMode: Text.Wrap
+                }
+
+                Label {
+                    width: parent.width
+                    id: minimumLatencyDesc
+                    text: qsTr("A lower bound on display latency. Set higher to account for more jitter in your connection, at the cost of increased display latency.")
+                    font.pointSize: 9
+                    wrapMode: Text.Wrap
+                }
+
+                Slider {
+                    id: minimumLatencySlider
+
+                    value: StreamingPreferences.minimumLatency
+
+                    stepSize: 1
+                    from : 0
+                    to: 50
+
+                    snapMode: "SnapOnRelease"
+                    width: Math.min(minimumLatencyDesc.implicitWidth, parent.width)
+
+                    onValueChanged: {
+                        minimumLatencyTitle.text = qsTr("Minimum latency: %1 msec").arg(value)
+                        StreamingPreferences.minimumLatency = value
+                    }
+
+                    Component.onCompleted: {
+                        // Refresh the text after translations change
+                        languageChanged.connect(onValueChanged)
+                    }
+                }
             }
         }
 
--- a/app/settings/streamingpreferences.cpp
+++ b/app/settings/streamingpreferences.cpp
@@ -45,6 +45,7 @@
 #define SER_CAPTURESYSKEYS "capturesyskeys"
 #define SER_KEEPAWAKE "keepawake"
 #define SER_LANGUAGE "language"
+#define SER_MINIMUMLATENCY "minimumlatency"
 
 #define CURRENT_DEFAULT_VER 2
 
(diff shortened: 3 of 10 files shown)


picarica:
how was this not implemented yet ?
great work

AzazKamaz:
I haven't checked the code yet but will share my own thoughts.

Ideally we should stabilise entire frame path from frame capture to frame display. In such a way we also compensate encoding and decoding time jitter.

As base timeline we can use frame capture timestamps (maybe need to send them from sunshine?) from which we can get delays between frames.

We could even get remote VRR with this!

kirksaunders:
@AzazKamaz
 
 > Ideally we should stabilise entire frame path from frame capture to frame display. In such a way we also compensate encoding and decoding time jitter.
> 
> As base timeline we can use frame capture timestamps (maybe need to send them from sunshine?) from which we can get delays between frames.

I'm a bit skeptical of such an end-to-end solution, since it's really hard to synchronize timestamps between two separate systems. You can't simply trust raw timestamps sent from a separate system (since their system clocks aren't synchronized), and you can't rely on an "ack" or "frame displayed" message, since that would include any network latency.

Definitely an intriguing idea, just not sure how it would look in practice. Perhaps someone more knowledgeable than me could make it work.

AzazKamaz:
> I'm a bit skeptical of such an end-to-end solution, since it's really hard to synchronize timestamps between two separate systems. You can't simply trust raw timestamps sent from a separate system (since their system clocks aren't synchronized), and you can't rely on an "ack" or "frame displayed" message, since that would include any network latency.

The thing here is that you actually don't need an absolute timestamps to be synchronised between two systems. You can use differences between incoming timestamps to know how to display frames on local system.

You basically keep delay as unknown but just do frame pacing. The only thing you have to choose is some kind of base delay offset: if you base you local pacing timeline just on the first arrived frame it is possible that a lot of frames will arrive later than needed. So it's essential to add some artificial delay.

For sure you can just set it to constant value of for example 100ms (after first frame arrival timestamp), it will give you the smoothest experience but also uncomfortable latency, so I guess the way to go here is to record timings of frames arrival and calculate what delay we need to get for example 99.9% of frames under it...

Such a solution of dynamic pacing latency may be an overkill for LAN streaming tho) in that case jitters and latency are not so big so constant (configurable) value may be ok

AzazKamaz:
I noticed that Parsec was smoother than moonlight last time I tried (with comparable latency). Do not really know what they are doing but as in commercial product they definitely put effort into this thing and also have statistics from around the world to tune it

AzazKamaz:
> The thing here is that you actually don't need an absolute timestamps to be synchronised between two systems. You can use differences between remote timestamps to know how to display frames on local system.

The thing we need here is that monotonic clock is monotonic and uniform which is not really always true (it seems so after checking rust docs for [std::time::Instant](https://doc.rust-lang.org/std/time/struct.Instant.html#monotonicity), also check [this answer on C++ comparison](https://stackoverflow.com/a/73304871))
And we definitely do not want to use just clock time because it gives jitters itself when it syncs.

freddyli:
Does this PR also help with audio crackling?

I have the issue that due to jitter (I think) audio is unstable. More so with the android client, but also on Windows as client. Server is Sunshine on Windows (AMD-encoding). Parsec is running perfectly fine audio 100 %. I have tried all the settings and work-arounds I have found on the internet.

ody:
Would love to see this get resurrected. Would be super to have for playing RPGs over WiFi

Deus-nsf:
I was priviledged to test the visual side of things (not sound yet) with a +50ms of buffer delay, which I think is a good tradeoff between queueing and still having a somewhat responsive gaming experience, and you know what?
It did work surpringly well! It won't magically fix a bad network but it can definitely smooth things out :)
+50ms is still less than google stadia delay so we should be fine hehehe

andygrundman:
After a discussion in Discord today, I ran some builds with the OP's patch (plus a couple of [minor fixes](https://github.com/moonlight-stream/moonlight-qt/compare/master...andygrundman:moonlight-qt:andyg.min-latency-test)) and they're available for a short while.
[Windows](https://ci.appveyor.com/project/andygrundman/moonlight-qt/build/job/a57knf2mc18t9gwe/artifacts)
[Mac](https://ci.appveyor.com/project/andygrundman/moonlight-qt/build/job/h475bw775gwqf94x/artifacts)
[Linux](https://ci.appveyor.com/project/andygrundman/moonlight-qt/build/job/vc5oed0s7tvr4bg6/artifacts)

kirksaunders:
It's cool to see some interest in this PR. I've largely quit using game streaming, which is why I haven't touched this PR or rebased it recently. But if others find it useful, maybe I can touch it up.

You're right that it doesn't buffer audio along with the video; that's definitely a miss. I'd also like to see a more quantitative evaluation of whether it improves pacing. The best I've done is monitoring the presentation frame time graphs from my steam deck.

mspeedo:
I would be really interesting in this. I got occasional micro stutters on my client (Lenovo Legion Go S) due to network jitter packet loss (0.55%) even on free 6GHz wifi channel. But when I switched from HW decoding (sub 1ms decode/render latency) to software decoding (5 + 5 = 10ms decode/render latency), these stutters were completely gone. I hope this will get eventually implemented.

oryschakj:
Definitely interested in checking this out, too, to see if it resolves my audio skip/micro stutter issues on macOS on an MBP M4 Max.

Karabacak021:
I would be really happy to test it, I just tried everything but did not find solution to these micro stutters: [https://www.youtube.com/watch?v=VbkSt60iWzY](https://www.youtube.com/watch?v=VbkSt60iWzY)
And I had the same idea in my head, that it would be cool to add 6-10ms delay to buffer.

AbrahamHuitron:
> I would be really happy to test it, I just tried everything but did not find solution to these micro stutters: https://www.youtube.com/watch?v=VbkSt60iWzY And I had the same idea in my head, that it would be cool to add 6-10ms delay to buffer.

This is the same stutter issue I have on my setup that I haven't been able to fix.
(9800x3d + 5090; Steam Deck docked w/ethernet.)

PR #1158, opened 2024-01-18
Add game mode with new UI

moi952:
Hello,

I added :

- adding a computer to favorites
- a gamepad in the header, it's a toggle that activates a "game mode"
- If game mode is activated moonlight launches in full screen and displays the apps of the favorite computer, if the computer is not turned on, it sends a WOL and once the computer is activated it displays the apps
- A palette in the header which allows you to change the size of application images with slide, allows you to change the background which is in gradient, allows you to change the background which is in gradient, spacing of application images, font size of app name
- I added bubble particles that move randomly

This information is recorded so each time it is launched the theme remains the same, the app is launched on the same computer.

Once game mode is deactivated, nothing changes compared to before, except that the applications are not displayed in grid form but in a horizontal scrolling list.

I still have work to do but I think we're already on a good V0.1

Capture d’écran 2024-01-18 à 21 54 04
Capture d’écran 2024-01-18 à 21 57 30
Capture d’écran 2024-01-18 à 21 53 33
Capture d’écran 2024-01-18 à 21 56 25
Capture d’écran 2024-01-18 à 21 53 42


paulohenriquefurtadopereira:
nice, much better than the standard native interface. It would be interesting to leave an animated video background in the PS4 and PS5 style by default, in addition to replacing the game loading with an animated loading. It's interesting to add a sound effect when selecting the app and not least a hover effect like pulsating.

PR #1182, opened 2024-02-12
SDL3 : WIP

boberfly:
Hello,

I went on an adventure to port moonlight-qt to SDL3. Now SDL3 isn't stable yet so I don't expect this to be merged anytime soon, however I put this PR up so it is known that this has been done and there isn't any duplicated efforts.

SDL2 is still supported with a lot of pre-processor checks. Linux-only on Wayland/X11 for now until other platforms are tested. I've made a best-effort conversion for the other platforms kmsdrm/EGL, windows/win32 but left macOS video decode alone, looks like that code is being worked upon right now.

Audio was a rewrite so this one needs checking with care to see if I'm doing the right thing.

My ultimate goal is to use the new `SDL_pen.h` API so I'll be working on this next.

Cheers, I'll set to draft and continue to force-push things as I discover them and do some testing.

### DIFF
--- a/app/streaming/audio/renderers/sdlaud.cpp
+++ b/app/streaming/audio/renderers/sdlaud.cpp
@@ -1,10 +1,18 @@
 #include "sdl.h"
 
 #include <Limelight.h>
+#if HAVE_SDL3
+#include <SDL3/SDL.h>
+#else
 #include <SDL.h>
+#endif
 
 SdlAudioRenderer::SdlAudioRenderer()
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    : m_AudioStream(nullptr),
+#else
     : m_AudioDevice(0),
+#endif
       m_AudioBuffer(nullptr)
 {
     SDL_assert(!SDL_WasInit(SDL_INIT_AUDIO));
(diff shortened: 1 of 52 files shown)


app/streaming/input/abstouch.cpp
```diff
@@ -30,13 +34,19 @@ Uint32 SdlInputHandler::longPressTimerCallback(Uint32, void*)
 
 void SdlInputHandler::disableTouchFeedback()
 {
+#ifdef Q_OS_WIN32
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    if (SDL_strcmp(SDL_GetCurrentVideoDriver(), "win32") == 0) {
```
icculus, line 39: "windows" not "win32"

app/streaming/audio/renderers/sdlaud.cpp
```diff
@@ -73,17 +113,26 @@ bool SdlAudioRenderer::prepareForPlayback(const OPUS_MULTISTREAM_CONFIGURATION*
...
     return true;
 }
 
 SdlAudioRenderer::~SdlAudioRenderer()
 {
-    if (m_AudioDevice != 0) {
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    if (m_AudioStream) {
+        // Stop and destroy audio stream
+        SDL_PauseAudioDevice(SDL_GetAudioStreamDevice(m_AudioStream));
```
icculus, line 126: These can be [SDL_PauseAudioStreamDevice](https://wiki.libsdl.org/SDL3/SDL_PauseAudioStreamDevice) to cut out the SDL_GetAudioStreamDevice call. There's an equivalent [SDL_ResumeAudioStreamDevice](https://wiki.libsdl.org/SDL3/SDL_ResumeAudioStreamDevice), too, for elsewhere.

While one _does_ need to resume to get a device from SDL_OpenAudioDeviceStream moving, pausing it before destroying the stream isn't necessary; it will halt playback just the same.

app/streaming/audio/renderers/sdlaud.cpp
```diff
@@ -102,7 +151,11 @@ bool SdlAudioRenderer::submitAudio(int bytesWritten)
 {
     // Our device may enter a permanent error status upon removal, so we need
     // to recreate the audio device to pick up the new default audio device.
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    if (SDL_AudioDevicePaused(SDL_GetAudioStreamDevice(m_AudioStream))) {
```
icculus, line 155: Just realized we never added a stream-specific equivalent of SDL_AudioDevicePaused (whoops!), so for now at least, this part _does_ need to call SDL_GetAudioStreamDevice().

app/streaming/audio/renderers/sdlaud.cpp
```diff
@@ -119,11 +172,22 @@ bool SdlAudioRenderer::submitAudio(int bytesWritten)
 
     // Provide backpressure on the queue to ensure too many frames don't build up
     // in SDL's audio queue.
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    // TODO: Check to see if this is correct and doing what SDL2 was doing.
```
icculus, line 176: This roughly matches SDL2, so this TODO can be removed.

app/streaming/input/abstouch.cpp
```diff
@@ -115,7 +137,23 @@ void SdlInputHandler::handleAbsoluteFingerEvent(SDL_TouchFingerEvent* event)
...
     if (LiGetHostFeatureFlags() & LI_FF_PEN_TOUCH_EVENTS) {
         bool isPen = false;
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+        int numTouchDevices = 0;
+        SDL_TouchID *touchId = SDL_GetTouchDevices(&numTouchDevices);
+        for (int i = 0; i < numTouchDevices; i++) {
+            if (event->touchId == touchId[i]) {
+                const char* touchName = SDL_GetTouchDeviceName(touchId[i]);
 
+                // SDL will report "pen" as the name of pen input devices on Windows.
```
icculus, line 147: Actual pen support (using Windows Ink for now, WinTab support coming later) just landed in SDL3, so if you want _actual_ pens and not touches, you should look for SDL_EVENT_PEN_* events.

Pens are a mess; in most cases we should be able to tell you all sorts of stuff about the hardware that's available, but most operating systems just sort of surprise you with the presence of a pen when it starts sending input events about it, and won't tell you anything _about_ the pen's abilities. As such, we built SDL's pen support to match that. Take a look at [the details](https://wiki.libsdl.org/SDL3/CategoryPen).

icculus (commented):
As requested [on Twitter](https://x.com/boberfly/status/1876520421236891728), I took a super-fast look at this (up to the gamepad parts) to provide feedback on the SDL3 integration details. If this was helpful/interesting, I can do a more thorough examination later on.

chenx-dust:
will this work continue?

PR #1190, opened 2024-02-18
Fix Discord icon on Qt 5

jorys-paulin:
This pull request is a hotfix for pull request #1183.

It turns out the new icon from that PR doesn't render properly if Moonlight is built with Qt 5 (which is used for releases):

Moonlight_2024-02-18_18-23-44

This pull request updates the SVG file to work with Qt 5 as well as Qt 6. It seems like the Qt 5 SVG renderer doesn't like when two zeroes in a path aren't separated by a space. I also added an explicit size to the SVG as well, just in case.

Had I tested my changes in the first PR with Qt 5 I would have probably caught it, so this time I made sure to check with both Qt 5 and Qt 6 (though only on Windows).


jorys-paulin:
Alright, I've checked the CI builds for Windows and Linux, everything looks as expected. Haven't check on macOS and Steam Link but I think this PR is ready for review and merging.

PR #1194, opened 2024-02-25
Replace material symbols with icons

jorys-paulin:
This pull request replaces the Material Symbols used in some parts of the UI with Material icons, used for every other icon, allowing for a consistant style of icons throughout the whole Moonlight UI.

The icons in question are the following:
- The warning icon used in the PC list when a PC is offline
- The play and stop icons used in the app list when an app is running

Here are screenshots of the changes:

warning-before-after
play-stop-before-after

I've also tried to update all icons to Material symbols to see how it would look, but ultimately decided to stick with Material icons, as those are used by basically every other Moonlight client, so I decided to stick with Material icons for consistency with those.
For reference, here's how the toolbar icon buttons would look like with Material symbols instead:

Moonlight_2024-02-12_18-29-26


rumplin (approved):

rumplin (approved):

cgutman:
I actually think the rounded corners of the symbols look better (I remember looking at both designs when I added these). If we going to change things, I'd prefer we go the other way and change them all to the rounded corner symbols.

jorys-paulin:
I've updated the icons to Material symbols.
Personnally, I don't think they fit with the rest of the interface, since they are rounded while other things like the box art isn't.
There's also a part of "it's been like this since the start in 2018" too, I have to admit.

Issue #1111, opened 2023-11-08
Provide Wake on LAN option to stream command


blackxored:
Given there’s wake on LAN functionality in Moonlight, I’d like to be able to invoke that logic from the command line as I have games that just run `Moonlight stream host game` from the launcher, but that doesn’t work when the PC is off.

I would think a `—wake` that tries to wake up the PC and then sleeps/retries till it’s on would be very helpful for people that share this use case.

PR #1203, opened 2024-03-07
Send WOL packet when seeking a computer.

hgaiser:
This PR makes it so that Moonlight sends a WOL packet when starting to seek for a computer. The goal was to allow a command like `moonlight stream <HOSTNAME> <APP>` to succeed, even if the host is asleep. Effectively, this PR fixes https://github.com/moonlight-stream/moonlight-qt/issues/1111.

I opted not to make the sending of the packet optional, since in my opinion there's no harm in sending it even if it's not required.

### DIFF
--- a/app/backend/computerseeker.cpp
+++ b/app/backend/computerseeker.cpp
@@ -6,6 +6,13 @@ ComputerSeeker::ComputerSeeker(ComputerManager *manager, QString computerName, Q
     : QObject(parent), m_ComputerManager(manager), m_ComputerName(computerName),
       m_TimeoutTimer(new QTimer(this))
 {
+    // If we know this computer, send a WOL packet to wake it up in case it is asleep.
+    for (NvComputer * computer: m_ComputerManager->getComputers()) {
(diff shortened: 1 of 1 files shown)


Nonary:
> I opted not to make the sending of the packet optional, since in my opinion there's no harm in sending it even if it's not required.


Well, other than people who have more than one host will suddenly start seeing all their devices wake up the moment they open Moonlight.


A better option would be have it sent when tapping on the offline instead, but really up to cgutman either way.

hgaiser:
Ahh I didn't know it was used there as well. Let me check that later and get back to you. One option is to add an argument that decides if WOL would be sent, which is true for the CLI and false otherwise. Thanks for the feedback.

cgutman:
I think I made the same mistake when reading this thinking it was ComputerManager rather than ComputerSeeker.

This codepath is indeed only used for the CLI case and it will only wake machines that match the name/UUID specified for this CLI operation, so this looks good.

blackxored:
Thank you so much for this PR! Been waiting for this for a while.

PR #1209, opened 2024-03-10
Add "show performance overlay" setting/option

jorys-paulin:
This pull request add a new streaming option and command line argument to automatically show the performance overlay when streaming starts. This feature is equivalent to the existing "Show performance stats while streaming" option of the Moonlight Android client.

The user can check the new "Show performance stats while streaming" option in the settings to enable this feature:

Moonlight_2024-03-10_21-01-55

They can also enable it via the command line by passing the new `--performance-overlay` option.

There is currently no checks to make sure the overlay is actually supported by the current platform (like it's the case with the Raspberry Pi and Steam Link), so the option can still be enabled on unsupported platforms, the overlay just won't appear.


moi952:
Hello, I see that your ticket is in WIP, the functionality is not completed?
Thank you for your work, hoping that it is merged

PR #1265, opened 2024-04-28
Fix Raised Black Levels on Steam Deck OLED

dylanfrankcom:
Added `getDecoderColorRange` to `plvk` and explicitly set the color range to full to fix raised black levels on OLED displays, an issue that has come up relating to #1117.

This seems to have fixed the elevated black levels when HDR is enabled in Moonlight for the Steam Deck OLED on my end. I don't have an LCD device to test this on so I'm not sure if it has a negative impact there.


cgutman:
Nice find!

PR #1305, opened 2024-06-06
Increase computer seek timeout to 30s.

hgaiser:
Now that https://github.com/moonlight-stream/moonlight-qt/pull/1203 is merged, moonlight tries to wake a computer when it wants to start streaming. However, the time for my computer to wake up and get a network connection is roughly 10s, so by the time a stream is successfully started, moonlight has already timed out.

This PR increases the timeout to 30s, which should be long enough for a computer to start streaming.

### DIFF
--- a/app/cli/quitstream.cpp
+++ b/app/cli/quitstream.cpp
@@ -7,7 +7,7 @@
 #include <QCoreApplication>
 #include <QTimer>
 
-#define COMPUTER_SEEK_TIMEOUT 10000
+#define COMPUTER_SEEK_TIMEOUT 30000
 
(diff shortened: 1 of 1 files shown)


cgutman:
This change is just to the seek timeout for the 'quit' action, which probably doesn't need to be longer for WoL (assuming the computer will probably _not_ be asleep while a game is running). I bumped it up to 30s for the 'stream' and 'list' actions in 486c46781b7f0ef62bdcb5ea178f95de264f6794

hgaiser:
Ah! I didn't know this was defined in multiple locations. Thanks!

PR #1377, opened 2024-08-03
BitrateUnlock

PlayXboxtion963:
Unlock bitrate upper to 1000Mbps which has been test in Apple M1+GTX1650. 
Picture quality is better than  150Mbps and network can be  about 70Mb/s


WinkelCode:
I haven't done any detailed picture quality comparison myself, but I can confirm that Apple's Media Engine is able to consume HEVC streams with this bitrate. I used it between a Nvidia RTX 4070 and Mac with an M2 Pro chip, the bandwidth is variable depending on what's on screen, but Little Snitch (macOS network monitor) reported around 80MB/s with motion on screen. I did not observe any glitchyness in the stream. There are probably diminishing returns at those bandwidths, but I am running an Ethernet cable directly between the devices, so might as well use what's available.

PR #1391, opened 2024-08-17
Fix #1350 viewing apps and streaming at Ubuntu Bionic L4T

### LINKED ISSUE #1350: Unable to start session due to QML bug
**Describe the bug**
After successfully adding a host PIN, I'm unable to start the connection (this is on v6.0 L4T; prior version worked).  Two error messages appear for every click, below:

Qt Warning: QQmlComponent: Component is not ready
Qt Warning: qrc:/gui/main.qml:82:5: QML StackView: push: nothing to push

**Steps to reproduce**
Install moonlight-qt on your Jetson Nano and try to connect.

**Other Moonlight clients**
Connections to the host from my other clients works fine.

**Client PC details (please complete the following information)**
 - OS: Ubuntu Bionic Beaver (final image available for the Jetson Nano from nVidia)
 - Moonlight Version: bionic, 6.0.1-1 arm64
 - GPU: Nvidia
 - Linux package type (if applicable): deb

**Server PC details (please complete the following information)**
 - OS: Gentoo Linux
 - Sunshine or GeForce Experience version: sunshine 0.23.1
 - GPU: Quadro RTX 4000
 - GPU driver: 555.58.02

gabrielformiga:
This PR fixes opening stream / apps on L4T Bionic as reported at #1350

### DIFF
--- a/app/gui/AppView.qml
+++ b/app/gui/AppView.qml
@@ -8,7 +8,7 @@ import SdlGamepadKeyNavigation 1.0
 
 CenteredGridView {
     property int computerIndex
-    property AppModel appModel : createModel()
+    property AppModel appModel
     property bool activated
     property bool showHiddenGames
     property bool showGames
@@ -27,6 +27,7 @@ CenteredGridView {
     }
 
     Component.onCompleted: {
+        appModel = createModel()
         // Don't show any highlighted item until interacting with them.
         // We do this here instead of onActivated to avoid losing the user's
         // selection when backing out of a different page of the app.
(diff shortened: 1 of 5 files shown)


cgutman:
It looks like doing creation this way breaks the parent relationship with the views:
```
00:00:08 - Qt Warning: qrc:/gui/AppView.qml:66: Error: Qt.createQmlObject(): Missing parent object
00:00:08 - Qt Warning: qrc:/gui/CenteredGridView.qml:10: TypeError: Cannot read property 'width' of null
```

gabrielformiga:
It's when we go back to PcView from AppView screen, creating objects like that the StackView takes the ownership of the created object and destroys it during stackView.pop(). That warning wasn't logged before because when we push by objects instead of url's, the pop action doesn't destroys the object and also sets the parent back to the originalParent. This warning is not really a problem, but if it that log annoys you I can try to find another solution.

gabrielformiga:
@cgutman, Fixed the warnings by moving function calls / property definitions that was depending of parent instance during initialization by moving it to Component.onCompleted. Also fixed a object lifecycle problem when transitioning from QuitSegue to StreamSegue. Everything os ok now, with no more nullpointer warnings. Let me know if there is something else to adjust, thanks.

PR #1399, opened 2024-08-28
Native CoreAudio renderer with spatial audio

andygrundman:
Hi! I'm just going to drop this here, but it should be considered alpha quality and nowhere near ready to be merged. But it does work quite well for me and I'd like to get it out there for some feedback and more eyes/ears on it. I am pretty impressed at how good it sounds and the fact that CoreAudio doesn't even break a sweat doing spatial processing on 8 channels of PCM. It would prefer to take 12 channels actually, but that's a problem for another day...

---
* Inspired by an example app from Apple. [1]
* Needs an M1 or newer Mac. Eventually should work on iOS/tvOS, although you might need iOS 18.
* Operates in a standard passthrough mode for stereo or when you have enough real channels (HDMI).
* When headphones or built-im Macbook speakers are detected, this enables spatial audio via the amazing AUSpatialMixer, which is capable of rendering any number of channels up to 7.1.4 Atmos in very high quality binaural stereo.
* Added a new section in the upper-right of the stats overlay with audio stats.

Planned features:
* Head-tracking and personalized HRTF, but since these are technically private user data, you need a $99 Apple developer account to build apps that include these features. I'm still in the process of doing that, and this will need more investigation on how to manage this in the real build.
* I want to add a few config options to customize a few of the settings, such as overriding a bad guess about your USB device being a headphone.

[1] https://developer.apple.com/documentation/audiotoolbox/generating_spatial_audio_from_a_multichannel_audio_stream


andygrundman:
I've posted a build of this for anyone who would like to try it out. Be sure to read the release notes. 
https://github.com/andygrundman/moonlight-qt/releases/tag/v6.0.2-sa1

Murmur2k:
Thank you Andy - came across this by accident today, super excited to see this in action - I'm on an M3 so hope to be able to test out and let you know my feedback.

andygrundman:
Please try with the newest build from https://github.com/andygrundman/moonlight-qt/releases which fixes a bug with the surround channels.

andygrundman:
> This is really great work Andy, thank you - I see your latest changes too for the MacOS end of the UI. Do you have a new release I can grab 

Thanks! I've uploaded a 3rd build, which adds a few options in settings:

spatial-audio-settings

Note about head-tracking: This is not a properly signed build, and my understanding of Apple's security is that the two spatial audio features that require entitlements will be prevented from working when you install an unofficial build.

If head-tracking works after you enable it, or if you see the text "Render mode: personalized spatial" in the overlay, please let me know.

Murmur2k:
> Note about head-tracking: This is not a properly signed build, and my understanding of Apple's security is that the two spatial audio features that require entitlements will be prevented from working when you install an unofficial build.
> 
> If head-tracking works after you enable it, or if you see the text "Render mode: personalized spatial" in the overlay, please let me know.

Just tried it out and head-tracking is working perfectly using AirPods Pro (Gen 2) and MacBook Pro M3. 
Audio
video

The only issue I'm having is infrequent audio drop outs but doesn't appear to be as a result of your build so I'll need to dig into that further - perhaps just wifi related albeit on 6E. I tested with some games (Mafia, Fort Solis as a couple examples), 7.1 test videos etc. The Mac is finally a killer portable gaming machine! Thank you

andygrundman:
> Just tried it out and head-tracking is working perfectly using AirPods Pro (Gen 2) and MacBook Pro M3. 

Well that's great news, I guess this is a beta entitlement and maybe they'll lock it down later. Hopefully they realized it's a bit silly to lock out head-tracking, knowing someone's "head pose" doesn't really feel like sensitive user data.

> The only issue I'm having is infrequent audio drop outs

Does it work better on ethernet? Does it work better if you only stream stereo audio? How about compared to the release version (7.1 should get downmixed by SDL in the release build) but use the same bitrate, I think.

I am wondering if we need to give the user more control over the audio bitrate and the 5ms packet size. The problem with this is that anything along these lines would need a Sunshine patch as well. I'm hoping it's just a bug in the way I'm handling the audio somewhere.

I admit most of my testing has been with ethernet, as I have some annoying wifi issues on my M1 Pro MBP as well as an M4 iPad using 6E, and have a hard time getting a smooth video stream without ethernet. It's the kind of thing that sometimes clears up if you disable Bluetooth or reboot, or there's always the chance it's my Unifi setup. The iPad is my only 6E device and it's real fast in a speedtest, but Moonlight video and audio both stutter a lot at any bitrate.

Murmur2k:
I have ordered an ethernet adaptor so will see if that improves things and report back.

Attached a small video in case it helps with the audio cutting out and the metrics showing:
https://github.com/user-attachments/assets/18eb2b9c-b2a2-4b34-9a13-5935d0652d79

Interestingly after some fiddling it appears to go away if I do not enable the "mute host speakers whilst streaming" option

Murmur2k:
> Well that's great news, I guess this is a beta entitlement and maybe they'll lock it down later. Hopefully they realized it's a bit silly to lock out head-tracking, knowing someone's "head pose" doesn't really feel like sensitive user data.

ah I should add that I'm on the dev 15.1 Beta (24B5035e)

Murmur2k:
Bit more digging here - ethernet does resolve so the audio stuttering is WiFi + MacOS related. Seeing lots of Mac related audio stuttering threads out there (e.g. https://github.com/moonlight-stream/moonlight-qt/issues/159) pointing to potential Mac services, none of the suggestions work for me.. its not bitrate/throughput related but I do see it occur when the blue overlay shows packets dropped due to resync >0%

Any thoughts/ideas as to how to resolve would be much appreciated as the beauty of Moonlight for me is to be portable

rramnel:
Thanks for implementing this feature. I am on the M1 Pro and it works with no issues with my AirPods Max. Head tracking also works. I am on wifi and I don't have any issues like audio stuttering.

rramnel:
I have iPadOS 18 Public Beta installed on my iPad Pro. Will Spatial Audio along with head tracking work on it? How can I get access to the iPadOS Beta version of Moonlight?

andygrundman:
> I have iPadOS 18 Public Beta installed on my iPad Pro. Will Spatial Audio along with head tracking work on it? How can I get access to the iPadOS Beta version of Moonlight?

So far this is only for macOS, but iOS/tvOS is definitely the next task.

ody:
Heads up. Merging this together with https://github.com/moonlight-stream/moonlight-qt/pull/1180 results in segfault when enabling the stats overlay. When I was initially testing 1180 I did so by merging the two PRs together and it worked fine but I noticed the author said that the enhancements on [macOS could add ~6ms](https://github.com/moonlight-stream/moonlight-qt/pull/1180#issuecomment-2471966838) but I was only seeing a 1ms difference so I went about remerging and building. I must have done something wrong the first time around because subsequent builds caused a segfault and no matter the order in which I merged the code together, it segfaulted.

The nature of how 1180 was kept current with master, by merging master into the branch makes it hard to rebase. I've had to merge these together by rebasing https://github.com/moonlight-stream/moonlight-qt/pull/1410 onto master then this one on top of that, then merge 1180 into the rebased 1399.

Sorry, I have not found the time to troubleshoot the segfault

momala454:
does it work on windows too ?

nortido:
Is there any chance to get this feature works?
Maybe some AI help?

Issue #1355, opened 2024-07-10
broken on Mac OS sequoia dev beta 3


frootloop88:
I know its a beta but a little heads doesn't hurt.

Got Mac OS sequoia beta 3 installed and I  get SecPKCS12Import failed: - 23000 (error99) when trying to pair. Beta 2 had no problems

HunterAhlquist:
I'm getting this just trying to connect.

HunterAhlquist:
I'm able to fix it by building the program from source.

frootloop88:
> I'm able to fix it by building the program from source.

How can I do this?

SwaMusic:
Can confirm. **SecPKCS12Import failed: -25300 (Error 99)** here as well!

andygrundman:
Rebuilding also worked for me. Here are the rough steps I used to build it, hope it's helpful for someone.

* Install Xcode 16 Beta from https://developer.apple.com/download/all/?q=Xcode
* Read over the Mac stuff in the readme, it boils down to:
* install qt using brew: `brew install qt --with-debug`
* Checkout all the code and build:

```
git clone https://github.com/moonlight-stream/moonlight-qt.git
cd moonlight-qt
git submodule update --init --recursive
qmake6 moonlight-qt.pro
make release
```

If all went well, the newly built app will be at `app/Moonlight.app` and you can just copy this over the broken one in the Applications folder.

frootloop88:
Maybe the dev can fix it and update it?

clastrum:
> Rebuilding also worked for me. Here are the rough steps I used to build it, hope it's helpful for someone.
> 
> * Install Xcode 16 Beta from https://developer.apple.com/download/all/?q=Xcode
> * Read over the Mac stuff in the readme, it boils down to:
> * install qt using brew: `brew install qt --with-debug`
> * Checkout all the code and build:
> 
> ```
> git clone https://github.com/moonlight-stream/moonlight-qt.git
> cd moonlight-qt
> git submodule update --init --recursive
> qmake6 moonlight-qt.pro
> make release
> ```
> 
> If all went well, the newly built app will be at `app/Moonlight.app` and you can just copy this over the broken one in the Applications folder.

As mentioned before from other guys, i encounter same issue after updating sequoia dev beta3. 

and I try these codes, and It works!!

Thank you for your kind solution.

brainvine:
Thanks for the fix. Can confirm: same issue here.

[...]

PR #1410, opened 2024-09-14
Support high-resolution stats; add audio stats framework

andygrundman:
NOTE: this patch depends on a patch to moonlight-common-c, see [this PR](https://github.com/moonlight-stream/moonlight-common-c/pull/95).

* Adds an audio stats overlay that works with all current renderers, showing common info such as bitrate and packet loss. It is blue and in the upper-right, and will appear whenever the video overlay is enabled.
* Audio renderers are able to add more lines to the overlay (the upcoming CoreAudio patch uses this).
* Added bitrate/FEC display to both video and audio stats.
* Consolidated the 3 FPS lines into one to save a bit of space.
* All time-based stats are now microsecond-based, improving accuracy of very fast events.

moonlight-stats


cgutman:
I haven't performed a detailed review yet, but it looks great at first glance. I will mark this for the v6.2 milestone, since v6.1 must ship in the next couple days to fix #1355 for Sequoia.

andygrundman:
Sounds good. I know large patches are a pain and I sometimes get a bit carried away, so I appreciate the review.

Dregu:
This layout is not ideal on small screens, should probably just go below video stats.

image

(And may I suggest adding the renderer platform to the video side too, since audio seems to have it. It's a bit of a hassle to figure out if you're running on Xwayland for example...)

cgutman:
@andygrundman can you rebase your branch on the latest master, switch back to the upstream moonlight-common-c submodule, and split the audio changes out of this PR?

Let's just take the video overlay improvements for v6.2. I want to think more about whether we want to expose video and audio stats with the same settings or separately.

andygrundman:
> @andygrundman can you rebase your branch on the latest master, switch back to the upstream moonlight-common-c submodule, and split the audio changes out of this PR?
> 
> Let's just take the video overlay improvements for v6.2. I want to think more about whether we want to expose video and audio stats with the same settings or separately.

This PR is quite old and I agree, it should focus only on the internals and updating the current stats' timing resolution. I'll get this done ASAP.

andygrundman:
Rebased, all audio stuff removed. I also updated the bandwidth calculation to use the same class I use on Xbox. I left the rest of the stats unchanged.

I also set `frame->pts` to the new rtpTimestamp value, and set the ffmpeg time_base to 90000. This should have no effect on anything but can be there for future use.

cgutman:
I gave this a try locally and the bitrate numbers seem quite a bit off from what Task Manager reports. I think much of the problem is that we're only measuring the encoded video bitrate across the wire, so we're leaving out FEC data in the calculation.

image

I wonder if it would be better to add fields to `RTP_AUDIO_STATS` and `RTP_VIDEO_STATS` to count the raw data coming in and use that for the bitrate reporting.

The high-resolution stats look good though. You can move the bitrate stuff into a separate PR and we can merge the improved stats now, or we can continue discussion here and land them both after the bitrate concerns are ironed out.

cgutman:
Since we need this to unblock the latest moonlight-common-c updates, I'm going to merge it and turn off the bitrate reporting until my feedback is resolved.

PR #1413, opened 2024-09-17
Fix build for Xcode < 14

MattTannahill:
Restores directives to avoid symbol `kCMVideoCodecType_AV1` when building with Xcode < 14.

These directives were removed by https://github.com/moonlight-stream/moonlight-qt/commit/eb6d16fdcf1cc3101f216573f9de6a7c49af87db.

### DIFF
--- a/app/streaming/video/ffmpeg-renderers/vt_base.mm
+++ b/app/streaming/video/ffmpeg-renderers/vt_base.mm
@@ -58,6 +58,7 @@ bool VTBaseRenderer::checkDecoderCapabilities(id<MTLDevice> device, PDECODER_PAR
         }
     }
     else if (params->videoFormat & VIDEO_FORMAT_MASK_AV1) {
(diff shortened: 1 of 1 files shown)


MattTannahill:
@cgutman I'm a near-daily user of Moonlight on MacOS.  Unfortunately, I'm using a 2014 MacBook that can't be updated past Big Sur 11.7.10 (I can't bring myself to get a new computer if this one hasn't bit the dust).  Due to being stuck on Big Sur, the latest Xcode I can use is 13.2.1.  This version of MacOS is no longer supported, but I'd like to be able to contribute to Moonlight, if possible.  I'm still able to build and run the client on my machine if these directives are restored.

cgutman:
Xcode 14 was meant to be the official requirement to build Moonlight v6.1 on macOS, but I forgot to update the readme to reflect that. I'm fine with accepting patches for unofficial Xcode 13 support (provided the changes required aren't a maintenance burden), but it may break periodically since nobody is testing it.

PR #1435, opened 2024-10-04
Add "View details" to applications list

jorys-paulin:
This pull request adds a new option to an application's context menu that displays its details, specifically its name, ID and whether it supports HDR streaming.

Moonlight_2024-10-04_13-09-39
Moonlight_2024-10-04_13-09-48

Like with pull request #1221, I've based the implementation on Moonlight Android, which already has that feature.

### DIFF
--- a/app/gui/AppView.qml
+++ b/app/gui/AppView.qml
@@ -335,6 +335,16 @@ CenteredGridView {
                     ToolTip.timeout: 5000
                     ToolTip.visible: hovered
                 }
+                NavigableMenuItem {
+                    parentMenu: appContextMenu
+                    text: qsTr("View Details")
+                    onTriggered: {
+                        showAppDetailsDialog.name = model.name
+                        showAppDetailsDialog.hdrSupported = model.hdrSupported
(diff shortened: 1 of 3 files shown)


jorys-paulin:
Sidenote: this pull request should be merged before #1194, because I need to update it so the new dialog uses the new icons.

cgutman:
I'm not sure this really makes sense to merge with so little information provided from the host. The HDR support details are only provided with GeForce Experience hosts and even then aren't very accurate. We've stopped relying on this indication for HDR-supporting apps in Moonlight for several years and just enable HDR regardless of what the host says about the app's support for it.

jorys-paulin:
We could parse more information from the app list, but most of that info is only provided by GeForce Experience hosts (I'm thinking of the `Distributor` field for example).
Should we add this kind of information or just close this pull request alltogether?

cgutman:
I'm going to close it for now. If Sunshine starts sending stuff that we want to display, we can revisit this. I don't think it's worth doing GFE-specific features at this point.

Issue #1254, opened 2024-04-11
Media keys not forwarded despite "capture in fullscreen" configured


levouh:
**Describe the bug**
Media keys (play/pause, volume up/down, etc.) not forwarded

**Steps to reproduce**
Configure Moonlight to capture keys in fullscreen, press "media keys", observe they apply to the host.

---

I am mostly interested in the source code that is used to capture keys on the host, and how it does filtering (if any) for which keys to send through and which to not. The information above is rather sparse as I'd like to take a crack at fixing this myself, but would need feedback from the maintainers on:

1. Whether this is a "won't fix" kind of thing, and if so, why
2. Where I would start

cgutman:
The problem with this is that some platforms (Windows and macOS) don't seem to allow applications to suppress media keys. The normal keyboard capture mechanisms that apply to other keyboard shortcuts don't work on media keys, so we might be able to send them to the remote PC but they would also 

> I am mostly interested in the source code that is used to capture keys on the host, and how it does filtering (if any) for which keys to send through and which to not. 

Moonlight itself doesn't filter the keys. We use [SDL2](https://github.com/libsdl-org/SDL/tree/SDL2)'s `SDL_SetWindowKeyboardGrab()` API which uses per-platform backends to handle capture. For example, it uses aWH_KEYBOARD_LL hook on Windows and the `keyboard-shortcuts-inhibit-unstable-v1` Wayland protocol on Linux.

levouh:
It seems like it would potentially be a worse user experience to send the key press to both the client and host system, right? At least without explicit consent from the user. And even then, I'm not sure something like having a program on the client "eat" the key press would work (as perhaps this would also prevent it from being registered by the client at all), but I would need to experiment with it.

I figured an easier approach, especially since I imagine I am in the minority as far as desiring this behavior goes, might be to just use something like AHK and `ControlSend` (which allows sending keys to a specific application) to "only" send the key press to Moonlight. It seems it gets ignored (host system sees no key press come through) presumably because SDL2 standing somewhere in between the emulated key press on the client and the host system.

Of course I can take things like F13-F24 and map those to deliver the correct key presses to the host system as well so it is not out of reach by any means. Do you think this is something even worth pursuing a "fix" for (as you noted, sending them to the host/remote PC)?

PR #1453, opened 2024-10-31
Add key codes for audio & multimedia control keys.

Bobbar:
Adds maps for volume/mute, play/pause and prev/next track.


cgutman:
I will have to think about this. I wrote my concerns about media keys here: https://github.com/moonlight-stream/moonlight-qt/issues/1254#issuecomment-2054121296

Bobbar:
Ah, I didn't notice that issue. My bad!

I hadn't considered the local effects on the client OS. I understand the conundrum now.

PR #1502, opened 2025-01-05
Update qml_ru.ts

kmnkv1:
Typo corrected


cgutman:
Thanks for the contribution. We use [Weblate](https://hosted.weblate.org/projects/moonlight/moonlight-qt/) rather than PRs for our translations, so please submit your change there and it will automatically be picked up for the next release.

PR #1537, opened 2025-02-20
Add "Show hardware acceleration warning" option

### LINKED ISSUE #1536: Provide an option to disable "no functioning hardware accelerated video decoder" messages
I use a steam deck with Windows. The graphics drivers will never support hardware acceleration on Windows. Moonlight works plenty well for me to stream the desktop of my computer in 4K. I do not need Moonlight to tell me **every. single. time.** I. open. the. app. that. hardware. acceleration. isn't. working. It will never be working, and nobody can do anything about it. Nonetheless, I am fine so long as I am not constantly told I need something I don't.
Seriously as often as I open moonlight to work on my other PC, it's frustrating to have to tell this message to bug off. I have already set "force software decoding" hoping it would stop notifying me, but it doesn't.

jorys-paulin:
This pull request fixes #1536 by adding an option in the settings to hide the hardware acceleration warning dialog that pops up when no hardware video decoders are found, like on the Steam Deck running Windows.


Dregu:
I think "force software decoding" should be enough to disable the hardware warning and adding options to disable individual warnings is overkill...

PR #1555, opened 2025-03-15
UI Optimization

qiin2333:
Performance Overlay: A more elegant performance monitor overlay with real-time bandwidth calculation display.

PCView: Assigned easily distinguishable icons to each added host; allows dragging images as custom backgrounds.

AppView: Uses the cover image of the currently streaming game as the AppView background; the cover image is correctly cropped, fixing the issue of stretching and distortion.

SettingView: The bitrate adjustment slider uses a logarithmic scale to optimize the experience of setting high and low bitrates (0-800mbps).

Fixed: The issue of missing buttons in QT6 dialogs has been resolved.


moi952:
Hi, it's more fun than the current version. I tried your version. It would be nice to be able to choose the background image in the settings.

I'm having trouble understanding the image count information in the statistics. I like this approach: https://github.com/Axixi2233/moonlight-qt/releases, with a very concise statistics version and a more detailed version.

In the settings, there's a back arrow that overlays the parameter text, as does the version text.

Capture d’écran 2025-03-16 à 09 17 33

PR #1561, opened 2025-03-19
DualSense adaptive trigger support

ABeltramo:
This is part of a series of PRs:

## Moonlight client
 - https://github.com/moonlight-stream/moonlight-common-c/pull/102 to extend the Moonlight protocol for this new event
 - https://github.com/moonlight-stream/moonlight-qt/pull/1561 requires the changes in `moonlight-common-c` and exposes the adaptive trigger events via SDL

## Sunshine server
 - https://github.com/games-on-whales/inputtino/pull/26 to expose the adaptive triggers events in Linux
 - https://github.com/LizardByte/Sunshine/pull/3738  requires the changes in `moonlight-common-c` and `inputtino` and delivers the adaptive trigger events to Moonlight clients

https://github.com/user-attachments/assets/c403a181-bae7-4b79-a5b6-7278f5bba97e

### DIFF
--- a/app/streaming/input/gamepad.cpp
+++ b/app/streaming/input/gamepad.cpp
@@ -900,6 +900,17 @@ void SdlInputHandler::setControllerLED(uint16_t controllerNumber, uint8_t r, uin
 #endif
 }
 
+void SdlInputHandler::setAdaptiveTriggers(uint16_t controllerNumber, DualSenseOutputReport *report){
+
+#if SDL_VERSION_ATLEAST(2, 0, 16)
+    if (controllerNumber <= MAX_GAMEPADS && m_GamepadState[controllerNumber].controller != nullptr) {
+        SDL_GameControllerSendEffect(m_GamepadState[controllerNumber].controller, report, sizeof(*report));
+    }
+#endif
+
+    SDL_free(report);
+}
+
 QString SdlInputHandler::getUnmappedGamepads()
 {
     QString ret;
--- a/app/streaming/input/gamepad.cpp
+++ b/app/streaming/input/gamepad.cpp
@@ -903,7 +903,12 @@ void SdlInputHandler::setControllerLED(uint16_t controllerNumber, uint8_t r, uin
 void SdlInputHandler::setAdaptiveTriggers(uint16_t controllerNumber, DualSenseOutputReport *report){
 
 #if SDL_VERSION_ATLEAST(2, 0, 16)
-    if (controllerNumber <= MAX_GAMEPADS && m_GamepadState[controllerNumber].controller != nullptr) {
+        // Make sure the controller number is within our supported count
+    if (controllerNumber <= MAX_GAMEPADS &&
+        // and we have a valid controller
+        m_GamepadState[controllerNumber].controller != nullptr &&
+        // and it's a PS5 controller
+        SDL_GameControllerGetType(m_GamepadState[controllerNumber].controller) == SDL_CONTROLLER_TYPE_PS5) {
         SDL_GameControllerSendEffect(m_GamepadState[controllerNumber].controller, report, sizeof(*report));
     }
 #endif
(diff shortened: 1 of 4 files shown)


moi952:
Hi, does this allow the DualSense to be detected as a DualSense for compatible games?

I think it's currently emulated as a PS4 controller, is that correct?

Is it only compatible with a Linux host? Or is it also compatible with Windows?

ABeltramo:
@moi952 the Moonlight code in this PR would send the adaptive trigger events to the locally connected joypad across platforms; I've tested it on Linux and OSX.  

Server side, only Linux has full support for a DualSense joypad currently.

moi952:
> @moi952 the Moonlight code in this PR would send the adaptive trigger events to the locally connected joypad across platforms; I've tested it on Linux and OSX.  
> 
> Server side, only Linux has full support for a DualSense joypad currently.

1 - Windows doesn't support it because it has to be developed on Sunshine? 
2 - What's changing? 

Sorry, these may be stupid questions. 

On Windows, directly on my PC, the DualSense works well, people would be happy if it worked. 

3 - Is it compatible for games that natively use DualSense?

Thanks

app/streaming/input/gamepad.cpp
```diff
@@ -900,6 +900,17 @@ void SdlInputHandler::setControllerLED(uint16_t controllerNumber, uint8_t r, uin
 #endif
 }
 
+void SdlInputHandler::setAdaptiveTriggers(uint16_t controllerNumber, DualSenseOutputReport *report){
+
+#if SDL_VERSION_ATLEAST(2, 0, 16)
+    if (controllerNumber <= MAX_GAMEPADS && m_GamepadState[controllerNumber].controller != nullptr) {
+        SDL_GameControllerSendEffect(m_GamepadState[controllerNumber].controller, report, sizeof(*report));
```
cgutman, line 907: We probably need to check if the gamepad is a PS5 controller before we send the output report to it.

ABeltramo: Thanks, makes sense. I've added a check for it

hgaiser:
I've been testing this through moonshine and it seems to be working great! With these PRs and Steam Input disabled, Horizon Forbidden West correctly sends trigger effects when firing arrows. Haven't tested any other games, but there's also very few games that use this functionality.

ABeltramo:
I've updated `moonlight-common-c` to bring in the merged changes, this should be ready for review.

If you need a quick way to test this out with a DualSense joypad you can use https://github.com/games-on-whales/sdl-jstest. After building, you can just call
```
./sdl2-jstest --gamecontroller 0
```
during a streaming session and by pressing the microphone button, it'll cycle through a couple of different trigger effects.  
(that is what I use on the video above :wink:)

moonlight-common-c/moonlight-common-c
cgutman, line 1: I think this is pointing to a commit in your fork instead. Head of the official repo is `e95feaf4951b8dc774671a5d6a1c31d76d78e3ac`.

cgutman: Actually it was correct prior to your last commit (7d6fe461939ae96d65941ef9286780720f68b160). If you drop that from this PR, it should build properly.

ABeltramo: Yep, I've definitely messed up when syncing my `moonlight-common-c` fork. Should be correct now.

daniele-athome:
Thanks for this! I'm going to try it soon. Does it by any chance work with Bluetooth too?

hgaiser:
Yep it does

daniele-athome:
I can confirm it works without Steam Input as if it's attached to the gaming host.

ABeltramo:
Nice, thanks for testing @daniele-athome!

Itay080:
Hi, thanks for doing this!
This may be a stupid question and I'm likely missing something but it doesn't seem to be working for me.

I've built moonlight-qt from source with the build-appimage script and copied it to my Steam Deck with a DualSense connected through Bluetooth.

Sunshine on my Linux host appears to detect and emulate the controller as a Dualsense but the adaptive triggers do not work in supported games 

Steam Input is disabled in both the client and the host

PR #1562, opened 2025-03-20
Add quit-app-and-exit shortcut Ctrl+Alt+Shift+E

jpoje:
Add a shortcut to quit the current streaming app and exit Moonlight entirely.

As an OSX Moonlight user it is awkward to quit out of a stream and the Moonlight app itself quickly.  As someone who does not use the "quit app on host pc after ending stream" setting because I do use detach sometimes, the action flow is: press "ctrl-alt-shift-q", click the quit app button, click confirm, then finally press "cmd-q".  This PR rolls those up into a single shortcut and it seems to work nicely.

I'm not 100% that the Qt quit() command is in the right place and catches all streaming cleanup work so do let me know if that's messed up.  The log looked fine.

As far as the key choice: due to existing shortcuts it's a tough decision.  In a vacuum I would, as a new user, expect something more similar to current browsers and other apps with commands like "-Q" to exit out of everything entirely and "-W" to close the current session.  However with "-Q" already taken for closing the current session I doubt there is motivation to change that. "-X" and "-D" are also taken so I went with "-E" for "exit" and being somewhat comfortable to hit.

(Also, I have added a setting to disable the quit-app-confirm popup in my fork because it drives me crazy - I could make a PR for that as well if there's interest)

### DIFF
--- a/app/streaming/input/input.cpp
+++ b/app/streaming/input/input.cpp
@@ -111,6 +111,11 @@ SdlInputHandler::SdlInputHandler(StreamingPreferences& prefs, int streamWidth, i
     m_SpecialKeyCombos[KeyComboTogglePointerRegionLock].scanCode = SDL_SCANCODE_L;
     m_SpecialKeyCombos[KeyComboTogglePointerRegionLock].enabled = true;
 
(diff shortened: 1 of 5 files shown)


moi952:
Hi! I think is not working on macOs

PR #1566, opened 2025-03-24
Add bitrate auto-adjustment toggle and reset button

MoreOrLessSoftware:
This change adds a 'Use Default' button next to the bitrate slider that resets the bitrate to the default value for the current resolution/FPS. It also implements an autoAdjustBitrate setting that controls whether the bitrate is automatically updated when resolution/FPS/YUV444 settings change.

- When the user manually adjusts the bitrate slider, autoAdjustBitrate is set to false
- When the user clicks the 'Use Default' button, autoAdjustBitrate is set to true
- Bitrate is only updated automatically when settings change if autoAdjustBitrate is true

### DIFF
--- a/app/gui/SettingsView.qml
+++ b/app/gui/SettingsView.qml
@@ -281,11 +281,13 @@ Flickable {
                                 StreamingPreferences.width = selectedWidth
                                 StreamingPreferences.height = selectedHeight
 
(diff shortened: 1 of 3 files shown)


cgutman:
Looks great, thanks

PR #1571, opened 2025-04-07
Migrate Win/Mac/Linux builds to Github Actions

andygrundman:
This is a port of all builds except Steam Link to GH Actions. Windows and Mac are more or less straight ports, on Linux I ran into trouble with the qt515 repo so I used an existing [Install Qt action](https://github.com/jurplel/install-qt-action/tree/v4) instead. This worked so well that I built both a 5.15.2 and 6.8.3 version for Linux. Both seem to work on my Steam Deck. 

* Yeah that is an older QT5 version (we had 5.15.4) but their tool only goes to 5.15.2 for some reason.  Could we distribute QT6 as the primary build with a backup if anyone needs it?
* Git branch and Git SHA are now part of the file metadata (in Info.plist on Mac, appdata.xml on Linux, Windows TBD).
* There is a weekly build of master that should run even if there are no changes.


andygrundman:
Turns out Ubuntu 20.04 runner is going away in a week, just my luck, so I will move this to 22.04.

andygrundman:
I had a thought, that since the GH Actions cache system is quite good, plus the fact that https://github.com/cgutman/moonlight-deps still depends on Appveyor, we could fairly easily re-integrate the building of these libs into the Win/Mac builds, just like how it works for the Linux builds.

PR #1580, opened 2025-04-21
Allow disabling pre-launch warning tooltips and delays

Odizinne:
This PR adds a new configuration option that allows users to **disable warning tooltips and their associated delays** before starting a stream.

Currently, Moonlight displays warning messages with a 3.5-second delay on every stream start. While these warnings are helpful for first-time users or when configuration issues exist, they can become repetitive and unnecessary for regular use. This change gives users the flexibility to disable these warnings when they're no longer needed.

I considered whether this should be merged with the connection quality warning setting, but decided to keep it separate since they serve different purposes. If preferred, we could combine these settings in the UI to avoid adding another option.

⚠️ Translations are not updated

### DIFF
--- a/app/gui/SettingsView.qml
+++ b/app/gui/SettingsView.qml
@@ -1237,6 +1237,17 @@ Flickable {
                 }
 
                 CheckBox {
+                    id: configurationWarningsCheck
+                    width: parent.width
+                    text: qsTr("Show configuration warnings")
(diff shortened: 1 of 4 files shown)


jorys-paulin:
Looking good! Don't worry about translations, it's fine

Odizinne:
I can do french translation if needed

PR #1584, opened 2025-04-27
Fix logger-based coredumps during early exits

FrogTheFrog:
On my linux laptop calling `moonlight --help` results in a crash/coredump after the help is printed.

It is caused by `00:00:00 - Qt Warning: QThreadStorage: entry 2 destroyed before end of thread 0x5555558b5c80` log entry that is being logged after `atexit(SDL_Quit());` (I don't know if that's the only log).

The crash occurs here:
```
void logToLoggerStream(QString& message)
{
    // ... omitted

    // Strip session encryption keys and IVs from the logs
    message.replace(k_RikeyRegex, "&rikey=REDACTED");
    message.replace(k_RikeyIdRegex, "&rikeyid=REDACTED");

    // ... omitted
}
```

The regex backend/engine used by Qt for `k_RikeyRegex` is already cleaned up and causes SEGFAULT at best, in other cases the process just hags for a minute or 2. Removing these `message.replace(...)` creates SEGFAULTs in other places.

This is my least invasive solution for suppressing the logging without refactoring global variables and thus preventing segfaults.

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -60,10 +60,88 @@
 // Log to console for debug Mac builds
 #endif
 
+// A helper class to ensure that all the logging callbacks have finished adding their logs to the
+// logger thread and ensures (via the help of the guard) that the callbacks can no longer log
+// after waitForAllLoggingToStop is called.
+class LoggerSuppressor final
+{
+public:
+    class SuppressorScopeGuard final
+    {
+    public:
+        SuppressorScopeGuard(LoggerSuppressor &parent) : m_Parent{parent.tryIncrementCounter() ? &parent : nullptr}
+        {
+        }
+
+        ~SuppressorScopeGuard()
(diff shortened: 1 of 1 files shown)


cgutman:
I don't think this would totally solve the issue because the `QMutex` in your static `s_LoggerSuppressor` could be destroyed before some later logger callback comes in and tries to acquire it. I took a simpler approach in 1bf86f52d368935e60780acc2d9d3c5265f6bce8. 

There's still a small race window if the log function fetched our logger callback address before we reverted it back to default and it hasn't yet queued a log message before the static variables start being destroyed. Unless we see evidence it's realistically possible to hit that race window, I'd like to avoid going crazy with synchronization (especially because the whole point of the logging thread is to avoid introducing delays in the threads trying to log).

PR #1585, opened 2025-04-27
Add custom timeout option for listing apps via cli

FrogTheFrog:
I would like to use `moonlist list <host>` cli command in my tool, but the default timeouts are too long if the host is offline.

I can terminate moonlight via signals after X seconds, but I would like to do it in the worst case scenario only. Thus, these new timeout overrides would be a nice addition.


FrogTheFrog:
Closing as no longer needed

PR #1587, opened 2025-04-28
Draft: add audio output device to settings

### LINKED ISSUE #1548: Allow selecting audio output device
**Is your feature request related to a problem? Please describe.**
When trying to set Moonlights output device in plasma to a audio output different than the system default, moonlight will immediately switch back to the default output device.

**Describe the solution you'd like**
If moonlight is that sticky with which device it binds to maybe add an option under audio settings to select the desired sink.

Renari:
This is my progress with adding an option for an audio output device:
image

There are currently are issues with this such as:
- It's currently only inside `sdlaud` I later realized there was multiple audio renderers and it should be moved to `IAudioRenderer`
- For some reason setting changes for this combo box in the front end are not being saved, I haven't noticed anything I'm doing differently than the current settings

resolves #1548


Renari:
I've moved these calls to `IAudioRenderer` however this now poses an issue with how I would extract a valid AudioRenderer to pull the device list, currently it seems all of this is done in Session, however inside of the UI here there is not a valid session.

I need to call `bool Session::initializeAudioRenderer()` without a session active, which would mean making a duplicate of this code that is callable from a static context.

Renari:
This is now completely setup for SDL2, however I'm not sure why this is failing:

```cpp
    if (!configuredDevice.isEmpty() && configuredDevice != "Default") {
        outputDevice = configuredDevice.toUtf8().data();
    }
    m_AudioDevice = SDL_OpenAudioDevice(outputDevice, 0, &want, &have, 0);
 ```

The audio device name is correct but this returns 0.

```
00:00:56 - SDL Error (0): Failed to open audio device: Invalid audio device instance ID
```

https://wiki.libsdl.org/SDL2/SDL_OpenAudioDevice

> a UTF-8 string reported by [SDL_GetAudioDeviceName](https://wiki.libsdl.org/SDL2/SDL_GetAudioDeviceName)() or a driver-specific name as appropriate. NULL requests the most reasonable default device.

Passing in this string should be valid.

Renari:
This is now working however I have no clue why:

```cpp
    const char* outputDevice = nullptr;
    QString configuredDevice;
    QByteArray utf8Data;
    if (Session::get() != nullptr) {
        // if there's not a valid session use the default audio device
        // this avoids a deadlock reading the settings before they are valid
        configuredDevice = StreamingPreferences::get()->audioDevice;
    }
    if (!configuredDevice.isEmpty() && configuredDevice != "Default") {
        utf8Data = configuredDevice.toUtf8();
        outputDevice = utf8Data.constData();
    }
    m_AudioDevice = SDL_OpenAudioDevice(outputDevice, 0, &want, &have, 0);
    if (m_AudioDevice == 0) {
        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION,
                     "Failed to open audio device: %s",
                     SDL_GetError());
        return false;
    }
    ```
   
This repeatedly fails, and eventually proceeds, maybe I need to setup `outputDevice` as a member variable so it does not go out of scope, I consistently get tons of logs about SDL failing to connect to the device and then after a minute or 2 it binds correctly.

Renari:
I tried using a member variable like so:
```cpp
    if (!configuredDevice.isEmpty() && configuredDevice != "Default") {
        delete m_OutputDevice;
        QByteArray utf8Data = configuredDevice.toUtf8();
        m_OutputDevice = new char[utf8Data.size() + 1];
        std::copy(utf8Data.constData(), utf8Data.constData() + utf8Data.size() + 1, m_OutputDevice);
    }
    m_AudioDevice = SDL_OpenAudioDevice(m_OutputDevice, 0, &want, &have, 0);
    ```
    This doesn't seem to change anything so I didn't commit it. I think the issue may be wayland permissions related, since I noticed it tends to fix itself once the window has focus.

Renari:
It looks like this was an upstream bug in SDL that was resolved here: https://github.com/libsdl-org/sdl2-compat/issues/367

So selecting audio devices in moonlight is no longer needed.

PR #1589, opened 2025-05-04
Add processing to resolve input issues on Japanese keyboard layout.

Haruka-127:
**Summary of Changes**

1. **Fixed unrecognized `\`/`_` key**
   In Japanese keyboard environments, Moonlight was unable to detect input from the key immediately to the left of the right Shift (the `\`/`_` key). This update ensures that those scancodes are now correctly recognized and processed.

2. **Corrected inverted Hankaku/Zenkaku key events**
   On Japanese keyboards, the key immediately to the left of `1` (the Hankaku/Zenkaku key) was generating its PRESSED and RELEASE events in reverse. We’ve added logic to invert those two event states so that PRESSED and RELEASE are reported correctly.

> **Note:** It’s not yet known whether these changes might conflict with other non-Japanese keyboard layouts.

---

**Test Environment**

* **Client:** Windows 11 24H2, Moonlight on Surface Go 4 (built-in keyboard)
* **Server:** Sunshine v2025.122.141614 on Windows 11 24H2

### DIFF
--- a/app/streaming/input/keyboard.cpp
+++ b/app/streaming/input/keyboard.cpp
@@ -413,7 +413,24 @@ void SdlInputHandler::handleKeyEvent(SDL_KeyboardEvent* event)
                 break;
             case SDL_SCANCODE_GRAVE:
                 keyCode = 0xC0;
-                break;
+                if (event->state == SDL_PRESSED) {
+                    SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION,
+                                "Grave key released (swapped)");
+                    LiSendKeyboardEvent2(0x8000 | keyCode,
+                                        KEY_ACTION_UP,
+                                        modifiers,
+                                        shouldNotConvertToScanCodeOnServer ? SS_KBE_FLAG_NON_NORMALIZED : 0);
+                    m_KeysDown.remove(keyCode);
+                } else {
+                    SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION,
+                                "Grave key pressed (swapped)");
+                    LiSendKeyboardEvent2(0x8000 | keyCode,
+                                        KEY_ACTION_DOWN,
+                                        modifiers,
+                                        shouldNotConvertToScanCodeOnServer ? SS_KBE_FLAG_NON_NORMALIZED : 0);
+                    m_KeysDown.insert(keyCode);
+                }
+                return;
             case SDL_SCANCODE_LEFTBRACKET:
                 keyCode = 0xDB;
                 break;
--- a/app/streaming/input/keyboard.cpp
+++ b/app/streaming/input/keyboard.cpp
@@ -428,9 +428,9 @@ void SdlInputHandler::handleKeyEvent(SDL_KeyboardEvent* event)
             case SDL_SCANCODE_APOSTROPHE:
                 keyCode = 0xDE;
                 break;
-            case SDL_SCANCODE_INTERNATIONAL1:
-                shouldNotConvertToScanCodeOnServer = true;
             case SDL_SCANCODE_NONUSBACKSLASH:
+                shouldNotConvertToScanCodeOnServer = true;
+            case SDL_SCANCODE_INTERNATIONAL1:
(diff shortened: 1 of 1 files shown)


app/streaming/input/keyboard.cpp
```diff
@@ -413,7 +413,24 @@ void SdlInputHandler::handleKeyEvent(SDL_KeyboardEvent* event)
...
+                                        shouldNotConvertToScanCodeOnServer ? SS_KBE_FLAG_NON_NORMALIZED : 0);
+                    m_KeysDown.remove(keyCode);
+                } else {
+                    SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION,
+                                "Grave key pressed (swapped)");
+                    LiSendKeyboardEvent2(0x8000 | keyCode,
+                                        KEY_ACTION_DOWN,
+                                        modifiers,
+                                        shouldNotConvertToScanCodeOnServer ? SS_KBE_FLAG_NON_NORMALIZED : 0);
+                    m_KeysDown.insert(keyCode);
```
cgutman, line 431: This seems like it might be a broader [SDL](https://github.com/libsdl-org/SDL) bug. Can you try one of the SDL test apps (ex: testgles2) with `SDL_EVENT_LOGGING=1` and see if you get reversed up/down events there too?

Haruka-127: I believe the issue lies in the fact that, on Japanese keyboards, when the Hankaku/Zenkaku (half-width/full-width) key is pressed, the OS issues a virtual key code that behaves differently from standard keys — it sends a "true" (released) event and immediately follows it with a "false" (pressed) event. (Scan code 41)


Here are the keyboard inputs obtained using a keylogger.

Datetime | Virtual Key Code | Scan Code | Extended Key Flag | Injected Flag | Context Code | Transition State Flag | Timestamp | Extra Info | Virtual Key Name | Display Name
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
07/06 12:02:44.407 | 65 | 30 | False | False | False | False | 298892765 | 0 | VK_A | A key
07/06 12:02:44.551 | 65 | 30 | False | False | False | True | 298892921 | 0 | VK_A | A key
07/06 12:02:44.947 | 66 | 48 | False | False | False | False | 298893312 | 0 | VK_B | B key
07/06 12:02:45.038 | 66 | 48 | False | False | False | True | 298893406 | 0 | VK_B | B key
07/06 12:02:46.342 | 67 | 46 | False | False | False | False | 298894703 | 0 | VK_C | C key
07/06 12:02:46.431 | 67 | 46 | False | False | False | True | 298894796 | 0 | VK_C | C key
07/06 12:02:47.435 | 244 | 41 | False | False | False | True | 298895796 | 0 | VK_OEM_ENLW | OEM specific
07/06 12:02:47.437 | 243 | 41 | False | False | False | False | 298895796 | 0 | VK_OEM_AUTO | OEM specific

cgutman: This is going to invert `SDL_SCANCODE_GRAVE` for all keyboard layouts though, which will break key layouts that handle `SDL_SCANCODE_GRAVE` properly. I think you should report this upstream to SDL where it can be handled in one place, rather than all SDL apps needing to work around this behavior.

cgutman (changes requested):
I'm not sure about these changes. They seem like they may adversely impact non-Japanese layouts and they seem to be workarounds for things that should be fixed in code outside of Moonlight (like SDL or Sunshine).

Haruka-127:
Thank you. I understand. I’ll try to find another solution.

ge9:
I personally worked around this by swapping "\\_" and "Hankaku/Zenkaku" with non-problematic and unused keys (like numpad keys) in the client and then re-swapping then in the remote machine. ([My Japanese blog post](https://turgenev.hatenablog.com/entry/2025/10/02/001505))
It seems like Moonlight "flattens" every keyboard input to English standard keyboard and so I suspect similar problems will happen in other non-English keyboards, but I don't have any confirmation.

PR #1591, opened 2025-05-07
add Meson build system support (linux OS is currently supported)

BruceKangCN:
QMake sucks, and makefile is quite slow. Meson can leverage the fast build speed of ninja and link speed of mold linker, and is much easier to set build options.


amazingfate:
I can drop [this pr](https://github.com/moonlight-stream/moonlight-qt/pull/1602) if we toggle to meson.

cgutman:
I appreciate the work, but I'm going to close this one. I think we'll just move to CMake in v7.0/v7.1 which has a Ninja generator.

PR #1605, opened 2025-05-28
Update SDL controller mapping submodule

pacwoodson:
Just pulled last version of https://github.com/mdqinc/SDL_GameControllerDB SDL game controller database.

Version linked is a bit outdated, missing a lot of new controllers. 

BTW, when there is a new version released ?


cgutman:
FWIW, Moonlight pulls updates dynamically from the submodule in https://github.com/moonlight-stream/moonlight-stream.github.io. I just updated that one, so clients should fetch the updated mapping database next time they start.

I enabled Dependabot to manage the submodule updates now, so I'll close this in favor of #1633

PR #1606, opened 2025-05-28
Added transparent background to overlays

joshrp:
Added settings to control text color (no longer fixed yellow / red) and background color and transparency.

This should help see the stats and messages while streaming.

Defaulted to white text on a 50% transparent background.

I'm not very familiar with this codebase or QT so I'd appreciate any other methods for doing this. The SDL part I'm pretty happy with and it should be very portable but maybe there's better calls to be using. 

Some previews:

image
image
image

### DIFF
--- a/app/gui/SettingsView.qml
+++ b/app/gui/SettingsView.qml
@@ -1268,6 +1268,121 @@ Flickable {
                     ToolTip.visible: hovered
                     ToolTip.text: qsTr("Prevents the screensaver from starting or the display from going to sleep while streaming.")
                 }
+            
+                Label {
+                    anchors.left: parent.left
+                    anchors.leftMargin: 10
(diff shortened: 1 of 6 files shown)


app/gui/SettingsView.qml
```diff
@@ -1738,7 +1853,7 @@ Flickable {
                     ToolTip.visible: hovered
                     ToolTip.text: qsTr("Display real-time stream performance information while streaming.") + "\n\n" +
                                   qsTr("You can toggle it at any time while streaming using Ctrl+Alt+Shift+S or Select+L1+R1+X.") + "\n\n" +
-                                  qsTr("The performance overlay is not supported on Steam Link or Raspberry Pi.")
```
joshrp, line 1741: I removed this as the stats overlay seems to be working fine on my Pi5 Bookworm

PR #1617, opened 2025-06-15
feat: disable text input when window gains focus

Qian23333:


cgutman:
Can you provide some context for this change? Text input status for an SDL window should persist across focus loss/gain, so I don't think this change should be neccessary.

ShadowLemoon:
image
The IME is annoying while streaming. I needed to change IME to English before streaming to avoid this situation.
This IME is from client, so we need to handle it on moonlight.

Coxxs:
I'm having the same issue. For reference, RDP (Remote Desktop app, pre-installed on Windows) automatically disables the IME when gains focus, which doesn't have this issue.

Update: Tried this PR, and it works fine for me.

cgutman:
This is almost certainly an SDL bug, but I'll take the fix anyway and hopefully we can revert it when we switch to SDL3.

Coxxs:
It seems that this fix no longer works after we switched to SDL3+SDL2-compat: an additional IME text box appears at the top left of the screen if the client (the PC running Moonlight) has IME enabled.
I tried creating a build at a commit that still uses SDL2 (https://github.com/moonlight-stream/moonlight-qt/commit/f4712718cbe802392b3649123820a205eb020032), and it doesn't have ime bug.

cgutman:
@Coxxs Thanks for the report. I fixed the issue in SDL3 and removed the now-ineffective workaround from this PR.

Once SDL 3.4.2 is released, I will update our prebuilt libraries which will fix this bug once more.

PR #1642, opened 2025-07-11
feat: Add custom window size option

hdkiller:
This change introduces a new feature that allows users to specify a custom window size when the display mode is set to 'Windowed'. This provides more flexibility for users who want to stream in a window that is not a standard resolution.

The following changes were made:
- Added `windowWidth` and `windowHeight` properties to the `StreamingPreferences` class to store the custom window size.
- Updated the settings UI to include input fields for the custom window dimensions. These controls are only visible when the display mode is set to 'Windowed'.
- Modified the `getWindowDimensions()` method in the `Session` class to apply the custom window size during the stream connection process.


andygrundman:
Looks great to me!

Issue #1681, opened 2025-09-01
High CPU usage when idle


cringegnere:
**Describe the bug**
Important CPU usage when Moonlight is idle.

**Steps to reproduce**
- Start Moonlight with a compositor (e.g. Weston), don't start any stream
- (Optional) Wait for HDMI device to go offline (sleep)
- Run top, process CPU usage is around 13% on my Machine (8c/16t). I assume it should be close to zero when idling

**Screenshots**
N/A

**Affected games**
N/A

**Other Moonlight clients**
N/A

**Moonlight settings (please complete the following information)**
- Tried with default settings and custom settings (e.g. keepawake, windowed, fullscreen, etc.), nothing changed

**Gamepad-related issues (please complete if problem is gamepad-related)**
N/A

**Client PC details (please complete the following information)**
 - OS: Debian 13 (6.12.41+deb13-amd64)
 - Moonlight Version: v6.1.0
 - GPU: Radeon 680M (from Ryzen 9 6900HX CPU)
 - Linux package type (if applicable): Tried AppImage and also local build

**Server PC details (please complete the following information)**
N/A

**Moonlight Logs (please attach)**

[moonlight.txt](https://github.com/user-attachments/files/22073032/moonlight.txt)

**Additional context**
- Same problem with Gamescope compositor
- Same problem running with DRM (no WM/Compositor)
- Compositor is run in embedded mode
- Could be related to HDMI device going offline (?)
- Display is set to 3840x2160@120.0 (if relevant)

eval-exec:
Hello, I create a PR tot try to fix the issue.

Could you help me to test? https://github.com/moonlight-stream/moonlight-qt/pull/1695

```bash
1. qmake6 moonlight-qt.pro
2. make -j20 release
3. ./app/moonlight    
```

cringegnere:
> Hello, I create a PR tot try to fix the issue.
> 
> Could you help me to test? [#1695](https://github.com/moonlight-stream/moonlight-qt/pull/1695)
> 
> 1. qmake6 moonlight-qt.pro
> 2. make -j20 release
> 3. ./app/moonlight

Sure, I'll try to test it in the following days, thanks!

cringegnere:
I've managed to do a quick test:

- Start moonlight
- Check cpu usage at idle -> approx 0%
- Launch a stream
- Quit the stream
- Check cpu usage at idle -> approx 12%
- Wait 15 min
- Check cpu usage at idle -> approx 12%

[...]

PR #1695, opened 2025-09-18
Reduce idle CPU by pausing hidden spinners

eval-exec:
## Summary
  - bind each BusyIndicator’s `running` property to `visible`
  - stop hidden spinners from scheduling frames when the UI is idle
  - avoids the steady render-loop churn that kept the process around 15% CPU

  ## Testing
  - launched `app/moonlight` after the change and observed idle CPU drop from ~15% to effectively 0%

### DIFF
--- a/app/gui/CliPair.qml
+++ b/app/gui/CliPair.qml
@@ -53,6 +53,7 @@ Item {
 
         BusyIndicator {
             id: stageSpinner
+            running: visible
         }
 
         Label {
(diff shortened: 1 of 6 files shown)


cgutman:
This is an _excellent_ find. I could never manage to track down why the render loop would never stop, even when the UI was not changing. This is also likely to help embedded devices without proper GL drivers which currently have very poor performance navigating the UI.

How did you figure out that it was the BusyIndicators causing the issue?

cringegnere:
Some feedback from my later tests https://github.com/moonlight-stream/moonlight-qt/issues/1681#issuecomment-3306790951

cgutman:
The post-streaming issue was because the `BusyIndicator` in `StreamSegue.qml` wasn't fixed. I addressed that in c9cb64f90b7ecfbb16dd8f287c833c023a2ea775.

PR #1712, opened 2025-10-20
build: Fix building on *BSD/riscv64

brad0:


cgutman:
Is there an equivalent mechanism to `hwprobe` on BSD that we should be using for ISA detection on RISC-V platforms?

brad0:
> Is there an equivalent mechanism to `hwprobe` on BSD that we should be using for ISA detection on RISC-V platforms?

I was going to look at adding FreeBSD/OpenBSD support for ARM and RISC-V after.

cgutman:
Ok sounds good.

Issue #1762, opened 2025-12-14
RPi 4/5 HDR Regression since 2023


praxis88:
**READ ME FIRST!**
If you're here because something basic is not working (like gamepad input, video, or similar), it's probably something specific to your setup, so make sure you've gone through the Troubleshooting Guide first: https://github.com/moonlight-stream/moonlight-docs/wiki/Troubleshooting

If you still have trouble with basic functionality after following the guide, join our Discord server where there are many other volunteers who can help (or direct you back here if it looks like a Moonlight bug after all). https://moonlight-stream.org/discord

**Describe the bug**
HDR doesnt work in gamestream for the vc4 driver anymore. This affects both rpi4 and rpi5 which both share the same raspbian OS. HDR mode on the TV turns off when gamestream starts, turns on when exiting the stream and returning to the QT app.

This may not have much at all to do with moonlight-qt, and may have been broken by a series of HDR "improvements" in the kernel for this vpu in 2023. This may be a kernel regression and vc4 regression.

**Steps to reproduce**
Fresh install, tick enable HDR.

**Screenshots**
Not really neccessary, HDR on vs off. 

**Affected games**
Not game specific

**Other Moonlight clients**
- Does the issue occur when using Moonlight on iOS or Android?

**Moonlight settings (please complete the following information)**
HDR on 1920x1080 20mbps HEVC hardware decode

**Gamepad-related issues (please complete if problem is gamepad-related)**
- Do you have any gamepads connected to your host PC directly?
- Does the problem still remain if you stream the desktop and use https://html5gamepad.com to test your gamepad?
  - Instructions for streaming the desktop can be found here: https://github.com/moonlight-stream/moonlight-docs/wiki/Setup-Guide

**Client PC details (please complete the following information)**
 - OS: Raspbian Trixie, Bookworm
 - Moonlight Version: 6.1.0
 - GPU: Broadcom vc4
 - Linux package type (if applicable): deb

**Server PC details (please complete the following information)**
Not really relevant. I have an MSI x86 miniPC that works properly with the same setup, same cables, same server, same tv, same everything. 

**Moonlight Logs (please attach)**
Nothing really relevant in the moonlight logs. Uses YCRCB encoding, BT2020_RGB HDMI colorspace

**Additional context**
[...]

PR #1765, opened 2025-12-19
Atomic modesetting for drm fixes HDR

praxis88:
Details here

https://github.com/praxis88/moonlight-qt/blob/master/HDR_IMPLEMENTATION_SUMMARY.md


praxis88:
https://github.com/moonlight-stream/moonlight-qt/issues/1762

This PR will not get merged, but this is implemented

PR #1775, opened 2026-01-01
metadata: Tidy up

yakushabb:
- Use desktop-application as the component type
- Drop categories that are already provided by the desktop file
- Group important information together
- Add vcs-browser URL to point to the source code repository

More information: https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines


cgutman:
Merged, thanks

PR #1780, opened 2026-01-07
feat: Profile support

distrohelena:
(code on this pull request was generated by Codex with my supervision)

**Reason**: I use Moonlight on my MacBook Air, where I constantly swap between working on my personal projects (where I want 16:10ish resolution at 60 FPS no HDR) and my TV (120 Hz 4K HDR).

**Feature**: Add a custom multi-profile combobox to the Settings menu, so you can change all settings per-profile.
image
As well as a quick change button on the main/pc view
image

### DIFF
--- a/app/backend/nvhttp.cpp
+++ b/app/backend/nvhttp.cpp
@@ -1,4 +1,5 @@
 #include "nvcomputer.h"
+#include "settings/streamingpreferences.h"
 #include <Limelight.h>
 
 #include <QDebug>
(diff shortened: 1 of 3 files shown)


moi952:
Oh thanks ! The best feature we need !!!!

distrohelena:
[moonlight-fork-profile.win-x64.zip](https://github.com/user-attachments/files/24541597/moonlight-fork-profile.win-x64.zip)

moi952:
This feature is awesome! Do you think it would be possible to associate a profile with a game?

We choose the default profile, and if we don't associate a specific profile with a game, it launches with the profile selected in the top bar.

For example, for a game that's quite demanding and I can't run it in 4K120, I could create a "4K60" profile and then associate it with games that are too demanding for my PC.

That would be amazing!

moi952:
@distrohelena , you responded to my message but didn't reply. 
Don't you find the idea interesting? 😜 
Linking profiles to apps. 

@cgutman , do you think this feature could be included in the next version?

distrohelena:
> @distrohelena , you responded to my message but didn't reply. Don't you find the idea interesting? 😜 Linking profiles to apps.
> 
> @cgutman , do you think this feature could be included in the next version?

I do! I think it might be a little bit more troublesome but should be okay to implement. I'm investigating it in a bit.

distrohelena:
My current branch has a _few_ more features at the moment, I'm pushing to another branch to avoid polluting this one, but I've added:
- Windowed Support: custom resolution pre-set on the settings (so it always opens at the specified resolution). Always show the mouse (streaming 60 Hz on a 240 Hz monitor looks awful, I need to make this an option but I rather see 2 cursors than watch the 60Hz one lagging).
- Custom Apollo branch: support for disableMainDisplay option, which my Apollo branch reads. Allows each client to disconnect or not my main display (so if I'm playing I can ask it to turn off my display, if not I can use my iPad as a secondary monitor without having to change settings).

moi952:
> > @distrohelena , you responded to my message but didn't reply. Don't you find the idea interesting? 😜 Linking profiles to apps.
> > @cgutman , do you think this feature could be included in the next version?
> 
> I do! I think it might be a little bit more troublesome but should be okay to implement. I'm investigating it in a bit.

That would be awesome! I'm on a Mac, so if you have a build with the profiles, I'd love to have it.

I checked your repo, and in the readme I tried to download the nightly build, but it doesn't include the profile update 😜

ns6089:
Good feature. It's not my place to speak for moonlight (yay/nay, time frame, etc), but if it's going to be merged, I believe server side should be made aware of these profiles. For this:
1. profiles should have persistent id in addition to a name
2. id-name pairs and active id should be passed to the server in `/serverinfo` https endpoint
https://github.com/moonlight-stream/moonlight-qt/blob/d4d4ab5b0686c7c1a5aae785b9a39cb615df5824/app/backend/nvhttp.cpp#L123
3. active profile id should be passed in `/launch` and `/resume` https endpoints
https://github.com/moonlight-stream/moonlight-qt/blob/d4d4ab5b0686c7c1a5aae785b9a39cb615df5824/app/backend/nvhttp.cpp#L189-L197

This way server can bind its own profiles to moonlight profiles, and moonlight won't need to add server-specific options to its config or bloat the UI.

PR #1785, opened 2026-01-13
feat: Add client-side cursor option for reduced input latency

TMCYTB:
# Adds client-side cursor option for a greatly reduced mouse input latency

## Problem

When streaming over high-latency connections (e.g., 150ms+), mouse input feels extremely sluggish because:
1. User moves mouse on client
2. Input is sent to host (~150ms)
3. Host moves cursor and renders frame
4. Frame is encoded and sent back to client (~150ms)
5. **Total: ~300ms delay before seeing cursor movement**

This makes the streaming experience nearly unusable for productivity work over long distances.

## Solution

This PR adds a **"Client-side cursor"** option that:
1. Tells the host to hide its cursor (make it transparent)
2. Shows the local system cursor on the client
3. Forces absolute mouse mode so positions stay synchronized

The result: **instant cursor response** regardless of network latency, while maintaining accurate click positions.

### Settings
- Added `clientSideCursor` boolean preference (default: false)
- Added checkbox in Settings UI with tooltip explaining the feature

### Protocol
- `nvhttp.cpp/h`: Send `clientSideCursor=1` parameter to host when launching session

### Input handling
- When `clientSideCursor` is enabled:
  - Force absolute mouse mode (ensures cursor positions stay synced)
  - Keep local cursor visible (don't let SDL hide it)
  - Skip SDL relative mouse mode (which hides the cursor)

## Files changed

- `app/settings/streamingpreferences.h/.cpp` - New setting
- `app/backend/nvhttp.h/.cpp` - Send parameter to host
- `app/streaming/session.cpp` - Pass preference to HTTP request
- `app/streaming/input/input.h/.cpp` - Handle cursor visibility and force absolute mode
- `app/gui/SettingsView.qml` - UI checkbox

## Companion PR

This feature requires host-side support. A companion PR to Apollo/Sunshine (https://github.com/ClassicOldSong/Apollo/pull/1336) adds:
- Transparent cursor replacement when client requests it
- Automatic cursor restoration on session end
- Crash recovery to prevent stuck invisible cursor

## Limitations

- **Games with custom cursors**: Games that render their own cursor won't benefit (their cursor is still in the video stream). This is primarily useful for desktop/productivity use.
- **Requires host support**: Only works with hosts that implement the `clientSideCursor` parameter (Apollo PR pending)

## Testing

Tested on Windows 11 host with ~150ms latency to client. Mouse feels instant while maintaining click accuracy, no more sloppy movements.

### DIFF
--- a/app/backend/nvhttp.cpp
+++ b/app/backend/nvhttp.cpp
@@ -194,6 +194,7 @@ NvHTTP::startApp(QString verb,
                  bool localAudio,
                  int gamepadMask,
                  bool persistGameControllersOnDisconnect,
+                 bool clientSideCursor,
                  QString& rtspSessionUrl)
 {
     int riKeyId;
@@ -223,6 +224,7 @@ NvHTTP::startApp(QString verb,
                                    "&remoteControllersBitmap="+QString::number(gamepadMask)+
                                    "&gcmap="+QString::number(gamepadMask)+
                                    "&gcpersist="+QString::number(persistGameControllersOnDisconnect ? 1 : 0)+
+                                   "&clientSideCursor="+QString::number(clientSideCursor ? 1 : 0)+
                                    LiGetLaunchUrlQueryParameters(),
                                    LAUNCH_TIMEOUT_MS);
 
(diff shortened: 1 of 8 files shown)


fragtion:
Why was this closed / not merged?

PR #1801, opened 2026-01-26
eglvid: Fix EGL initialization failure on Wayland

royka1:
The 'dummy renderer' hack, intended to force SDL_WINDOW_OPENGL flags on older platforms, causes the underlying Wayland surface or EGL state to become invalid on modern drivers (specifically Panfrost/Panthor).

This resulted in a black screen and 'Can't create shader: 0' errors because the EGL context could not be made current after the dummy renderer was destroyed.

This patch prevents the dummy renderer creation/destruction cycle when running under the Wayland video driver, fixing hardware decoding on Mali GPUs.

### DIFF
--- a/app/streaming/video/ffmpeg-renderers/eglvid.cpp
+++ b/app/streaming/video/ffmpeg-renderers/eglvid.cpp
@@ -441,33 +441,43 @@ bool EGLRenderer::initialize(PDECODER_PARAMETERS params)
         return false;
     }
 
-    // This will load OpenGL ES and convert our window to SDL_WINDOW_OPENGL if necessary
-    SDL_Renderer* dummyRenderer = SDL_CreateRenderer(m_Window, renderIndex, SDL_RENDERER_ACCELERATED);
-    if (dummyRenderer) {
-        SDL_DestroyRenderer(dummyRenderer);
-        dummyRenderer = nullptr;
-    }
-    else {
-        // Print the error here (before it gets clobbered), but ensure that we flush window
-        // events just in case SDL re-created the window before eventually failing.
-        EGL_LOG(Error, "SDL_CreateRenderer() failed: %s", SDL_GetError());
-    }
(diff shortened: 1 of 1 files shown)


cgutman:
We definitely can't skip the renderer initialization broadly like this because it will prevent the GLES renderer from working when Moonlight is built with libplacebo/Vulkan.

Under SDL, a window is either GL (`SDL_WINDOW_OPENGL`),  Vulkan (`SDL_WINDOW_VULKAN`), or neither. If you try to create a GL context for a window that doesn't have `SDL_WINDOW_OPENGL` set, it will fail. The only public way to turn a window from one type to another is to create a renderer (which invokes the private `SDL_RecreateWindow()` function behind the scenes), which is why we have this dummy renderer hack.

By selecting the `opengles2` renderer specifically, it also ensures the window has a GLES context rather than a desktop GL context (necessary because we use `OES_EGL_image_external`).

Can you provide some more details about the hardware and OS version you're running where you're seeing this issue? 

> This resulted in a black screen and 'Can't create shader: 0' errors because the EGL context could not be made current after the dummy renderer was destroyed.

Does reverting 30274f3ae824f6b4cc11ed5183bdf2a36920cd69 fix it?

royka1:
Thanks for your reply. I understand, unfortunately did reverting that commit not fix the issue. Perhaps I should make it that it only will be skipped when it's built without libplacebo/Vulkan?
The issue is with the Orange Pi 6 which works with the Panfrost mesa driver. It has a mvx decoder which is v4l2 stateful.
ffmpeg version 8.0.1: https://github.com/royka1/ffmpeg-sky1
Linux 6.18.4: https://github.com/royka1/linux
Debian Trixie

Logs:
[moonlight-working.txt](https://github.com/user-attachments/files/24917519/moonlight-working.txt)
[moonlight-not-working.txt](https://github.com/user-attachments/files/24917520/moonlight-not-working.txt)

PR #1838, opened 2026-03-07
fix(ffmpeg): try more times when receive EAGAIN during init

GenshinImpactStarts:
Some decoders require some time to produce the first decoded frame. However, commit [66a30c6](https://github.com/moonlight-stream/moonlight-qt/commit/66a30c66f3f2534f551f3b2e0f4c91c6751ab67a#diff-b079c7f05ea0ed57f958aff5a148e3d7f7959f7bff4243949adfcca74d3f50cfL674-L700) removed the retry mechanism when `avcodec_receive_frame` returns `EAGAIN` during the `FFmpegVideoDecoder` initialization stage.

This PR reintroduces the retry mechanism to avoid missing potential hardware acceleration opportunities.

I reused the retry count and delay duration that were used before that commit. I'm not sure whether the delay might be too long.

I observed this issue on the Spacemit K1 using the PowerVR driver.

### DIFF
--- a/app/streaming/video/ffmpeg.cpp
+++ b/app/streaming/video/ffmpeg.cpp
@@ -689,23 +689,33 @@ bool FFmpegVideoDecoder::completeInitialization(const AVCodec* decoder, enum AVP
             return false;
         }
 
-        err = avcodec_receive_frame(m_VideoDecoderCtx, frame);
-        if (err == 0) {
-            // Allow the renderer to do any validation it wants on this frame
-            if (!m_FrontendRenderer->testRenderFrame(frame)) {
+        // Some decoders require some time to output the first frame, so we'll
+        // retry a few times if we get an EAGAIN error.
+        for (int retries = 0; retries < 5; retries++) {
+            err = avcodec_receive_frame(m_VideoDecoderCtx, frame);
+            if (err == 0) {
+                // Allow the renderer to do any validation it wants on this frame
+                if (!m_FrontendRenderer->testRenderFrame(frame)) {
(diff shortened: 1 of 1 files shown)


cgutman:
I reverted that commit. Can you try `master` again and see if the issue is resolved?

GenshinImpactStarts:
Yes, it's resolved now. If everything looks good on your end, please feel free to close this PR.

GenshinImpactStarts:
BTW, I’m not very familiar with codec internals, so I might be missing some context here. I’d really appreciate your guidance on a few points:

- How significant are the benefits of sending EOS to the decoder?
- I noticed that the reverted commit mentioned: *"It causes hangs on defective v4l2m2m decoders, and it's not always well-supported by out-of-tree decoders either."*
  Does sending multiple frames instead of EOS avoid those hangs?
- Have there been real-world cases where decoders do not properly support EOS after only one frame?

I’d appreciate any clarification—thanks!

cgutman:
> How significant are the benefits of sending EOS to the decoder?

Well it's technically the only way you're guaranteed to get output, since decoders are allowed to buffer frames and not output them until additional frames are provided. However, in practice, some decoders are still broken and actually behave worse in the EOS case.

> Does sending multiple frames instead of EOS avoid those hangs? Have there been real-world cases where decoders do not properly support EOS after only one frame?

Yes, because v4l2m2m triggers an indefinite wait for output frames when you send EOS. The trouble is that some v4l2 drivers _still_ don't output the frame after EOS either. FFmpeg waits forever for the pending output frame and Moonlight deadlocks.

Those decoders may also fail to output frames when sending several frames as we do now, but at least we don't deadlock in that case.

PR #1857, opened 2026-04-10
Fix: DX11 not rendering on Snapdragon GPUs

linckosz:
### Bug description
I am using a Galaxy Book Go running Windows on ARM with a Snapdragon 7c Gen 2.

In production mode, streaming with H.264 works correctly. However, when switching to HEVC, the screen turns green and the decoder crashes after a few seconds.

In Debug mode, the application reports that the hardware is not supported and falls back to SDL rendering.

---

### Root cause
Snapdragon GPUs do not support Debug mode (`D3D11_CREATE_DEVICE_DEBUG` and `DXGI_CREATE_FACTORY_DEBUG`). This issue only affects developers.

Additionally, when using HEVC, Snapdragon GPUs do not support shared textures (`D3D11_RESOURCE_MISC_SHARED | D3D11_RESOURCE_MISC_SHARED_NTHANDLE`). This impacts end users, resulting in a green screen as shown below:

DX11-Bug

---

#### Debug mode
Attempt to create the device and DXGI factory with debug flags enabled. If it fails, retry without `D3D11_CREATE_DEVICE_DEBUG` and `DXGI_CREATE_FACTORY_DEBUG`.

#### Shared textures
Disable the use of a separate device for the decoder and renderer when running on Snapdragon with HEVC on Windows ARM.

---

### Result
Rendering now works as expected:

DX11-Fix


linckosz:
Artifacts are available for download here:
https://github.com/linckosz/moonlight-qt/actions/runs/24212413977#artifacts

cgutman:
> Snapdragon GPUs do not support Debug mode (D3D11_CREATE_DEVICE_DEBUG and DXGI_CREATE_FACTORY_DEBUG). This issue only affects developers.

This is not true. DXGI/D3D11 debug works fine on my Snapdragon X Elite system. Are you sure you have installed `Graphics Tools` as the [readme says](https://github.com/moonlight-stream/moonlight-qt?tab=readme-ov-file#windows-build-requirements)?

> In production mode, streaming with H.264 works correctly. However, when switching to HEVC, the screen turns green and the decoder crashes after a few seconds.

I cannot confirm this on my Snapdragon X Elite system (driver: 31.0.148.0) on Win11 25H2, but I've got a few other Snapdragon devices I will try to see if any of them are affected.

What's your GPU driver version and Windows version?

cgutman:
Okay, I was able to confirm on my QC710 (driver: 27.20.810.0) with Win11 22H2 that there's definitely a driver bug. It affects H.264 actually worse than HEVC in my case, preventing the decoder from initializing at all. I also see an initial green screen with HEVC, but it eventually recovers into a state where it renders properly but performance is extremely bad.

Unfortunately, using separate decoding and rendering devices cuts 3D engine usage at 4K60 by about 35% on the Snapdragon X Elite, so we are leaving some performance on the table by blocking this. However, given that X Elite has plenty of 3D headroom anyway (11% vs 17% in my test), I think this is fine until/unless we can come up with a more accurate check.

Thanks for finding this!

linckosz:
I only acquired an old Galaxy Book Go (Snapdragon 7c), don't have a Elite X to test it. I guess that if we would like to go further with the condition, we may distinguish more recent GPU by its feature.

PR #1872, opened 2026-05-01
Disable automatic mouse grab in windowed mode, add setting to opt-in

crazyallstar39:
By default, a new streaming session immediately grabs the mouse pointer on initialization. In windowed mode this is undesirable. You can't reposition the session window before the cursor gets captured, forcing a tab-out each time. This causes unnecessary fatigue.

This change makes windowed mode skip automatic mouse grab by default. Users can opt-in to grab in windowed mode via a new setting.

**Changes:**
- **`StreamingPreferences`**: Added `allowMouseGrabInWindowed` property (default `false`). When `true`, the session will grab the mouse even in windowed mode.
- **`Session::exec()`**: Extended the `needsPostDecoderCreationCapture` logic to skip grab in windowed mode unless `allowMouseGrabInWindowed` is enabled.
- **`SettingsView.qml`**: Added a checkbox "Grab the mouse pointer in windowed mode" in the Input Settings section.

## Test plan

- [ ] Build and launch the app, start a stream in windowed mode
- [ ] **Default behavior (unchecked)**: Verify the mouse pointer is NOT grabbed — you can click outside the window and move the window freely
- [ ] **Windowed mode with grab enabled**: Check the checkbox, reconnect the stream, verify the mouse pointer is captured within the application window
- [ ] **Fullscreen/fdesktop modes (regardless of setting)**: Verify mouse capture still works automatically in fullscreen modes
- [ ] **Persistence**: Toggle the checkbox, restart the app, verify the setting persists correctly

### DIFF
--- a/app/gui/SettingsView.qml
+++ b/app/gui/SettingsView.qml
@@ -1279,6 +1279,21 @@ Flickable {
                     ToolTip.visible: hovered
                     ToolTip.text: qsTr("Prevents the screensaver from starting or the display from going to sleep while streaming.")
                 }
+                CheckBox {
+                    id: mouseGrabCheck
+                    width: parent.width
+                    text: qsTr("Grab the mouse pointer in windowed mode")
+                    font.pointSize: 12
+                    checked: StreamingPreferences.allowMouseGrabInWindowed
+                    onCheckedChanged: {
+                        StreamingPreferences.allowMouseGrabInWindowed = checked
+                    }
+
+                    ToolTip.delay: 1000
+                    ToolTip.timeout: 5000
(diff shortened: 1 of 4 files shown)


cgutman:
I like this idea. I think the use cases for initial grab in windowed mode are minimal (and clicking into the window to grab is so simple), so I just implemented that behavior without a configuration option.

crazyallstar39:
Right on! Makes total sense. Happy to contribute to the project.

iclonethefirst:
Hey, I still would love to disable mouse grab entirely. I stream to a friend who open it on their secondary monitor and their mouse gets grabbed all the time.

PR #1909, opened 2026-06-12
Add novideo mode

Azkali:
Let user pass audio and input to the moonlight client device without video stream

### DIFF
--- a/app/gui/SettingsView.qml
+++ b/app/gui/SettingsView.qml
@@ -845,6 +845,22 @@ Flickable {
                     ToolTip.visible: hovered
                     ToolTip.text: qsTr("Frame pacing reduces micro-stutter by delaying frames that come in too early")
                 }
+
+                CheckBox {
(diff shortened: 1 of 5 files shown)


Azkali:
Any feedback is welcome, ideally I would like to prevent FFMpeg from running, it seems it is still running although not decoding anything, might be harmless. Secondly I would like to add text on a grey background using SDL just to inform user that they are in novice mode

jorys-paulin:
Nice feature, it's effectivly a "gamepad" mode. Having something shown while in this mode would be nice indeed.

Azkali:
> Nice feature, it's effectivly a "gamepad" mode. Having something shown while in this mode would be nice indeed.

Thank you much appreciated ! I did push some additional commits that show the app banner and name as well as a text on grey backgound. I will rename to gamepad mode otherwise it is ready to be merged on my end.

PR #1931, opened 2026-07-12
macOS: Add support for GIP (Xbox One/Series) controllers via direct USB bridge

viincnt:
## Summary

Adds support for GIP-protocol controllers (Xbox One/Series-generation
accessories - Microsoft-licensed pads from 8BitDo, PowerA, PDP, Razer,
and 20+ other vendors) on macOS, by reading them directly over USB and
feeding them into SDL as a virtual game controller.

macOS has no HID class driver for these controllers, so they're
invisible to SDL/the OS out of the box. The usual fix - creating a
system-wide virtual HID device - requires the
`com.apple.developer.hid.virtual.device` entitlement, which Apple
restricts to virtualization software vendors and isn't available here.
SDL2's virtual joystick API sidesteps this entirely: it registers a
joystick inside the process that creates it, without touching the OS
HID stack at all.

## How it works

`GipBridgeController` scans for known GIP vendor IDs (matching the
Linux `xone` driver's device table), claims the raw USB interrupt
IN/OUT endpoints via `IOUSBHost` (no privileged access needed since
no system driver has bound to these devices), and speaks GIP directly:
power-on handshake, LED handshake, input report parsing, rumble
(including the host-forwarded case with a safety-net timeout), and the
guide/Xbox button (delivered separately via `GIP_CMD_VIRTUAL_KEY`
rather than the regular input report). It feeds a virtual
`SDL_GameController` from a background queue; the existing
`SdlInputHandler` pipeline picks it up exactly like any other gamepad,
with no changes needed there.

This is entirely macOS-only, gated behind `Q_OS_DARWIN` in
input.h/input.cpp and a `macx { }` qmake scope in app.pro - it doesn't
touch Windows or Linux input handling at all.

## Notes for reviewers

- `gipbridge.mm`/`gipusbdevice.mm` compile under manual reference
  counting (MRC), matching this project's existing `.mm` files
  (`vt_metal.mm`, `vt_avsamplelayer.mm`) rather than introducing ARC,
  which would require touching those files' explicit
  release/autorelease calls. `QMAKE_OBJECTIVE_CFLAGS += -fobjc-weak`
  is added purely so the new files' async `IOUSBHost` callbacks can
  use `__weak self` captures (avoiding a retain cycle / a callback
  firing on a freed object if a device disconnects mid-I/O), without
  switching the whole project to ARC.
- GIP packet sequence numbers matter: a repeated sequence is treated
  by the device as a retransmission and silently dropped. This bit
  rumble specifically - a "stop" command reusing an earlier sequence
  could be ignored, leaving motors running - so rumble sequence
  numbers now increment on every call.
- Also implements `GIP_CMD_ACKNOWLEDGE` support for any incoming
  packet that requests one (a device that asks for an ack and never
  gets one keeps retransmitting). Not exercised by the hardware this
  was tested against, but part of the documented protocol.

## Testing

Hardware-tested end to end with an 8BitDo Ultimate Wired Controller
for Xbox (VID `0x2dc8`) streaming from a Windows GameStream/Sunshine
host: sticks, triggers, face buttons, d-pad, bumpers, stick clicks,
guide button, and rumble (including host-forwarded rumble and its
automatic timeout) all verified working.

Known cosmetic limitation, not addressed here: the controller's LED
stays solid after the stream ends instead of returning to its
blinking "searching for connection" pattern. Tried the documented GIP
LED blink/fade modes and none had a visible effect on this hardware;
not pursued further since it doesn't affect functionality.


GuiDev1994:
Great! I'm very interested in how it will be improved...
I have one Flydigi Apex 4, Gamesir G7 Pro, 8BitDo Ultimate and Ultimate C2 to test.

PR #1942, opened 2026-07-19
Add Fedora rpm build via packit

LecrisUT:
It would be quite helpful if you could add the rpm build automation with [packit](https://packit.dev/docs/guide) which would track the build status as this project and the Fedora dependencies evolve. The setup that is setup here is:
- Build a copr build for each PR
- If a build fails notify me to see the logs and try and help (would not mind having manual pings though :slightly_smiling_face:)
- `.distro/post-upstream-clone.py` fixes the `commitN` references so that it points to the git submodule shas

In order to enable this you would need to install the [github app](https://github.com/marketplace/packit-as-a-service).

A few other things that can be setup with this integration:
- Build nightly and releases into specific copr projects
- Add complex tests even some that would need setting up sunshine server and such

### DIFF
--- a/.distro/moonlight-qt.spec
+++ b/.distro/moonlight-qt.spec
@@ -1,0 +1,119 @@
+%global         forgeurl0 https://github.com/moonlight-stream/moonlight-qt
+
+# TODO: Debundle these once the build-system can do that
+%global         forgeurl1 https://github.com/moonlight-stream/moonlight-common-c
+%global         forgeurl2 https://github.com/cgutman/qmdnsengine
+%global         forgeurl3 https://github.com/gabomdq/SDL_GameControllerDB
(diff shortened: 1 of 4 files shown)


cgutman:
Are you sure that this will work for PRs since you've hardcoded the `https://github.com/moonlight-stream/moonlight-qt` URL?

LecrisUT:
> Are you sure that this will work for PRs since you've hardcoded the `https://github.com/moonlight-stream/moonlight-qt` URL?

Fair observation, yes it will. The way packit works is that it will edit the spec file and use a git archive of the current repo as `Source0` so it will even work in the local repo (you can try it in a container using `packit srpm` and decompress that to see the exact sources.

It is also why the spec version is written to be explicitly unbuildable in the current form (points to nonexistent version for example), instead needing to be fixed by packit first.

Issue #1789, opened 2026-01-17
The "?" "/" key from ABNT2 (Brazilian Portuguese) Keyboard is never sent from client to host (Unhandled button event: 135)


thedaryen:
**Description**
When using a Brazilian Portuguese keyboard on the moonlight client, the last key from the last letter row, which is "/" (or "?" with the shift modifier) is not sent to the host. I've enabled verbose logging on Sunshine and, while looking at it, keyboard packet events are created when any key is pressed, but not for the aforementioned key. While looking on Moonlight's log, it spams "Unhandled button event: 135" whenever that key is pressed.

I believe it might have to do with the fact that this particular layout has one additional key on the third letter row when compared to the US standard. 

<img width="830" height="722" alt="Image" src="https://github.com/user-attachments/assets/88103a88-9d06-4b93-af90-7ebd0f681ed6" />

The additional key on the brazilian layout is the ";" key right after >, but that one works ok. It's the "/" one that doesn't work on Moonlight because physically it's on a slot that doesn't exist for US keyboards (our right shift key is actually physically smaller to accomodate the aditional key).

It's an important issue to squash because this layout is the default for every standard brazilian-made keyboard, and the key is also pretty important for desktop use, both for writing and coding.

**Steps to reproduce**
Just try to use any standard Brazilian ABNT2 keyboard on the moonlight client. The "?" key won't be passed to the host. Changing settings like disabling or enabling always send scan codes doesn't change this behavior. Other keys work correctly (even "ç" for example).

**Other Moonlight clients**
This has been tested only on Windows, both for the client and for the Sunshine host (both latest version)

**Moonlight settings (please complete the following information)**
- All input settings related to Keyboard have been tried on both Sunshine and Moonlight. Changing the host or the client keyboard layouts on Windows also doesn't solve the issue, they just reflect the layout change, making all special keys wrong rather than just not sending the described key.

**Client PC details (please complete the following information)**
 - OS: Windows 11 25H2
 - Moonlight Version: v1.6.0

**Server PC details (please complete the following information)**
 - OS: Windows 11 25H2
 - Sunshine v2025.924.154138

**Moonlight Logs (please attach)**
- The only relevant log entry is "- SDL Info (0): Unhandled button event: 135" whenever the key is pressed

**Additional context**
Maybe related to: #1395

thedaryen:
[...]

PR #1532, opened 2025-02-11
Add some keys handling for support more keys on Japanese Keyboard Layout

rinsuki:
Fix #890 .

Added these keys:
* `¥` key, next to `^` key
* `_` key, next to `.` key
* Muhenkan (無変換) / Henkan (変換) keys

since I couldn't map first two keys to US keyboard layout, I needed to use SS_KBE_FLAG_NON_NORMALIZED flag, which is only available on Sunshine.

I'm not sure about this conflicts with other languages keyboard layout, it might be represents wrong character, but I think its still good than  got nothing when press key I guess.

Tested environment:

Moonlight on macOS 15.3 w/ MacBook Pro Internal Keyboard
Sunshine 2025.122.141614 w/ Windows 11 24H2

Note, currently this isn't properly works with Linux version of Sunshine, because it ignores SS_KBE_FLAG_NON_NORMALIZED flag.

PR #1944, opened 2026-07-20
Fix ABNT2 keyboard mappings for C1 and C2 keys

hertzoli:
## Summary

This PR fixes Brazilian ABNT2 keyboard handling by sending the ABNT-specific Windows virtual key codes for both affected SDL scancodes:

- `SDL_SCANCODE_INTERNATIONAL1` now maps to `VK_ABNT_C1`.
- `SDL_SCANCODE_KP_COMMA` now maps to `VK_ABNT_C2`.
- Both keys are sent with `SS_KBE_FLAG_NON_NORMALIZED` so the server does not convert them back through a generic scancode path.
- Fallback definitions are added for `VK_ABNT_C1` and `VK_ABNT_C2` when the build SDK headers do not define them.

## Why

On ABNT2 PT-BR keyboards, these keys are layout-specific and should be sent as their ABNT virtual key codes. Previously, `SDL_SCANCODE_INTERNATIONAL1` fell through to `SDL_SCANCODE_NONUSBACKSLASH` and used `0xE2`, while `SDL_SCANCODE_KP_COMMA` used `0x6C` instead of `VK_ABNT_C2`.

This is important because fixes that only handle `VK_ABNT_C1` leave the ABNT2 keypad comma key unresolved. This PR covers both `VK_ABNT_C1` and `VK_ABNT_C2`.

## Testing

- Tested locally with an ABNT2 PT-BR keyboard in the original commit this change was split from.
- Verified this PR diff only changes `app/streaming/input/keyboard.cpp`.

### DIFF
--- a/app/streaming/input/keyboard.cpp
+++ b/app/streaming/input/keyboard.cpp
@@ -6,6 +6,14 @@
 #define VK_0 0x30
 #define VK_A 0x41
 
+#ifndef VK_ABNT_C1
+#define VK_ABNT_C1 0xC1
+#endif
+
+#ifndef VK_ABNT_C2
+#define VK_ABNT_C2 0xC2
+#endif
+
 // These are real Windows VK_* codes
 #ifndef VK_F1
 #define VK_F1 0x70
@@ -337,7 +345,8 @@ void SdlInputHandler::handleKeyEvent(SDL_KeyboardEvent* event)
                 keyCode = 0x6B;
                 break;
             case SDL_SCANCODE_KP_COMMA:
-                keyCode = 0x6C;
+                shouldNotConvertToScanCodeOnServer = true;
+                keyCode = VK_ABNT_C2;
                 break;
             case SDL_SCANCODE_KP_MINUS:
                 keyCode = 0x6D;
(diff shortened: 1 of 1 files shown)


hertzoli:
This PR also fixes the Brazilian Portuguese (ABNT2) keyboard mapping issue for the two keys shown below:

image

 * `ABNT_C1` (the key next to the right Shift, `/ ? °`)
 * `ABNT_C2` (the keypad ABNT key `.`)

As a consequence, this also fixes **#1789**.

MLAN1O:
Tested this PR on my setup (Moonlight 6.1.0 on Fedora → Sunshine on Windows, ABNT2 PT-BR keyboard). Built from the agent/fix-abnt2-keyboard-mapping branch and can confirm the /? key (VK_ABNT_C1) now works correctly. This fixes issue #1789. Thanks for the fix! :)

cgutman:
I think this will regress the fixes by @rinsuki for Japanese keyboard layouts in #1532. Both send the same scancodes for different VK values.

Maybe we can check the `sym` value to disambiguate them?

arthuralvesdev:
Works with me, thank you so much for your work on it!!

PR #1950, opened 2026-07-23
Extend macOS stream content into the title bar

lwmacct:
## Summary

- Extend windowed macOS streams behind a transparent title bar while preserving the native window controls.
- Add a macOS-only setting to enable or disable the full-size content view.
- Reapply the native window style after fullscreen transitions or renderer-driven window recreation.
- Keep the title-bar region draggable while input capture is inactive.

## Testing

- Built the release target on Linux ARM64 with Qt 6.10.2 and SDL 2.32.10.
- Built and manually tested the feature on macOS.
- Verified that windowed stream content extends behind the transparent title bar while the native window controls remain available.

## Screenshot

Windowed macOS stream with content extending behind the transparent title bar


lwmacct:
@cgutman This is ready for review. It adds a macOS-only full-size content view for windowed streams, has been built and manually tested on macOS, and also passes a Linux ARM64 release build. The upstream CI has not started yet, so it may need maintainer approval for this fork PR. Thanks!

PR #1971, opened 2026-08-11
Fix SDL audio recovery after device failure

aaravrav:
`SdlAudioRenderer::submitAudio()` may report success when the SDL audio device has stopped, when the queued audio does not drain within the existing timeout, or when `SDL_QueueAudio()` fails. Reporting success in these cases prevents
Moonlight’s existing audio renderer recovery path from recreating the failed renderer.

This PR checks the SDL audio device status before accepting a submission, treats a queue that remains above the existing 50 ms threshold as a failure, and propagates errors from `SDL_QueueAudio()`. It also checks Moonlight’s pending audio duration only after confirming that the SDL device and its queue are healthy.


cgutman:
Can you explain what specific user scenario this fixes? I would like to reproduce the broken scenario locally for testing.

PR #1976, opened 2026-08-14
[Feat] Introduce Bluetooth Controller for no desktop installations

Drackrath:
- Introduced this for Devices directly running Moonlight on Startup
- Streaming Devices without Mouse & Keyboard input might need to scan for new Bluetooth devices

- Code written manually
- QML Files written with AI (since I have no clue how to do that)
- Refactored, sorted and debugged with AI
- Builds succeed
- Tested on Raspberry Pi Bookworm and Arch linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,9 @@ for different architectures, which handle building deps and extra linking for yo
     * To build from the command line for development use on macOS or Linux, run `qmake6 moonlight-qt.pro` then `make debug` or `make release`.
         * The final binary will be placed in `app/moonlight`.
     * To create an embedded build for a single-purpose device, use `qmake6 "CONFIG+=embedded" moonlight-qt.pro` and build normally.
-        * This build will lack windowed mode, Discord/Help links, and other features that don't make sense on an embedded device.
+        * This build will lack windowed mode, Discord/Help links, and other features that don't make sense on an embedded device. 
+        * On Linux, it additionally gains a BlueZ-backed Bluetooth pairing view for devices with no desktop environment to pair a controller from. 
+        Add `"CONFIG+=disable-bluez"` to leave it out.
         * For platforms with poor GPU performance, add `"CONFIG+=gpuslow"` to prefer direct KMSDRM rendering over GL/Vulkan renderers. Direct KMSDRM rendering can use dedicated YUV/RGB conversion and scaling hardware rather than slower GPU shaders for these operations.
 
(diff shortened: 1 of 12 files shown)


jorys-paulin:
Your feature look good to me, the QML seems to be perfectly correct as well. I should actually build your changes and see how the UI looks.

Since you seem to target embedded devices, maybe this code could be scoped to the `embedded` build? I know this is already done but I don't know how well QML integrates with that.

Drackrath:
@jorys-paulin I would try to create a new PR on the moonlight-embedded. Though I still needed the feature on the qt version due to issues with embedded in the past. Might try that version out first, whether it works in my setup and then we can scope it down. 

For usability, I'd keep this functionality, as it also enhances other systems which don't have a mouse & keyboard connected (Steam Deck, other Handhelds etc.)

jorys-paulin:
Sorry, I didn't mean Moonlight Embedded, Moonlight PC has an embedded build mode, as described in the README:

> create an embedded build for a single-purpose device, use qmake6 "CONFIG+=embedded" moonlight-qt.pro and build normally.
This build will lack windowed mode, Discord/Help links, and other features that don't make sense on an embedded device.
For platforms with poor GPU performance, add "CONFIG+=gpuslow" to prefer direct KMSDRM rendering over GL/Vulkan renderers. Direct KMSDRM rendering can use dedicated YUV/RGB conversion and scaling hardware rather than slower GPU shaders for these operations.

I think it's only used for the Steam Link build right now however, the Raspberry Pi version doesn't use it so it's not really worth it.

Drackrath:
Ah sorry, I misunderstood... Yeah the scope to the embedded build makes sense (I actually didnt use that on my pi), Will make necessary adjustments.

jorys-paulin:
I mean you don't have to, it's just that not a lot of people use Moonlight on the Steam Link compared to the Raspberry Pi, and your feature would be so nice when running Moonlight on RPI OS Lite. With the Link discontinued, even Valve recommends the Raspberry Pi with RPI OS (Lite or Regular) for a DIY streaming box.

Edit: Also, the Steam Link OS has settings for Bluetooth, but not RPI OS Lite (if I remember correctly), so your feature would be really useful on RPI OS Lite.

Drackrath:
It's already done. And yes I also run Raspi OS Lite with Sunshine & Heroic.

PR #1977, opened 2026-08-14
Add a context menu option to stream an app in a new window

### LINKED ISSUE #1205: Multiple sessions
I use macos and i copy moonlight.app to moonlight1.app to get multiple sessions, but a new window is a more elegant solutions.

yassinsolim:
## Problem

moonlight-common-c maintains a single global connection per process — `LiStopConnection()` and `LiInterruptConnection()` take no connection handle. Streaming two hosts concurrently therefore requires a second process, which users currently achieve by duplicating the app bundle (#1205).

## Change

Adds a **Launch in New Window** item to the app context menu that spawns a second instance already streaming the selected app.

- macOS uses `open -n` so LaunchServices starts a genuine second instance; other platforms re-exec `QCoreApplication::applicationFilePath()`.
- Passes the host UUID rather than an address, so it is stable across DHCP/VPN address changes.
- Gated on `SystemProperties.hasDesktopEnvironment`, matching existing usage in `SettingsView.qml`, so it does not appear on Steam Link or embedded builds.

## macOS Spaces interaction

`Session::initialize()` disables `SDL_HINT_VIDEO_MAC_FULLSCREEN_SPACES` when the configured resolution exactly matches a notched display's native mode. Left alone, both streams land on the same Space, which defeats the purpose of a second window.

The new instance is asked for the safe area resolution in that case only. Values come from `StreamUtils::getNativeDesktopMode()` at runtime — nothing is hardcoded. Non-notched displays, Intel Macs (the heuristic is inside `#if TARGET_CPU_ARM64`), and any non-native resolution are unaffected, and the whole block is inside `#ifdef Q_OS_DARWIN` so other platforms do not compile it. Command line options do not persist: `StreamingPreferences::save()` is only called from the settings screen.

Happy to drop this hunk if you would rather the client not adjust resolution implicitly. The feature works without it; separate Spaces on notched Macs is what is lost.

## Testing

Verified on macOS (arm64, notched display) against a Sunshine host, built with Qt 6.11.1:

- Builds clean; menu item renders and is correctly hidden without a desktop environment.
- Triggering it spawns a second process with the expected argv and that process connects and streams (`Video stream is 3024x1890x60`).
- The safe area substitution fires only in the intended case. With the saved native resolution the log reports `Overriding default fullscreen mode for native fullscreen resolution` (hint 0); with the substitution it reports `Overriding default fullscreen mode for native safe area resolution` (hint 1).

Not verified, stated plainly:

- **Two concurrent streams have not been observed running side by side on separate Spaces.** The mechanism is verified per-component, not end to end.
- **Windows, Linux and Steam Link are compile-only.** The `#else` path has not been exercised at runtime on any of them.

Feedback welcome on both gaps before this is considered mergeable.

Fixes #1205

### DIFF
--- a/app/gui/AppView.qml
+++ b/app/gui/AppView.qml
@@ -5,6 +5,7 @@ import QtQuick.Controls.Material 2.2
 import AppModel 1.0
 import ComputerManager 1.0
 import SdlGamepadKeyNavigation 1.0
+import SystemProperties 1.0
 
 CenteredGridView {
     property int computerIndex
@@ -298,6 +299,16 @@ CenteredGridView {
                 NavigableMenuItem {
                     text: model.running ? qsTr("Resume Game") : qsTr("Launch Game")
                     onTriggered: launchOrResumeSelectedApp(true)
+                }
+                NavigableMenuItem {
+                    text: qsTr("Launch in New Window")
+                    onTriggered: appModel.launchAppInNewInstance(model.index)
+                    visible: SystemProperties.hasDesktopEnvironment
+
+                    ToolTip.text: qsTr("Stream this app in a second Moonlight window, so you can stay connected to another host at the same time.")
+                    ToolTip.delay: 1000
+                    ToolTip.timeout: 5000
+                    ToolTip.visible: hovered
                 }
                 NavigableMenuItem {
                     text: qsTr("Quit Game")
--- a/app/gui/appmodel.cpp
+++ b/app/gui/appmodel.cpp
@@ -1,5 +1,12 @@
 #include "appmodel.h"
 
+#include "settings/streamingpreferences.h"
+#include "streaming/streamutils.h"
+
+#include <QCoreApplication>
+#include <QDir>
+#include <QProcess>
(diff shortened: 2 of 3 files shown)


jorys-paulin:
I'll leave the code review to the maintainer, but I don't see anything obviously wrong with your code.

The Steam Link build might be an issue, as an embedded device without a window manager, I'm not sure how launching another process with another window would be handled.

Also so you know, you can create (or switch) a pull request to be a draft if you want to clearly mark it as work-in-progress.

yassinsolim:
Sounds good.

The PR should be ready for review, it has everything you need to be able to run two instances of moonlight at the same time by right clicking the application on the host and clicking "Launch in New Window". This works on ARM-based macs so far. I believe it should work via command line for other computer architectures.

jorys-paulin (approved):

yassinsolim:
Hey @jorys-paulin do you know how to get the workflow approved? Is there a specific person that needs to run it?

Issue #1980, opened 2026-08-18
Steam Controller (2nd gen, 2026) is emulated as a DualShock 4


Drackrath:
Setup: 
moonlight-qt 6.1.0/master on Raspberry Pi 4, SDL 3.4.14, Steam Controller 2026 over its wireless dongle, Sunshine host with gamepad = auto.

Problem: 
During a stream the host emulates a DualShock 4 instead of an Xbox 360 pad, so games show PS glyphs and the pad behaves like a DS4.

Cause: 
SDL has no gamepad type for Steam Controllers, so gamepad.cpp sends LI_CTYPE_UNKNOWN while still advertising gyro/accel/touchpad capabilities. Sunshine's auto selection treats an unknown pad with motion/touchpad caps as a DualShock.

Fix: In the default: case of the type switch, report Valve controllers as Xbox:

default:
    // SDL has no type for Steam Controllers; they arrive as UNKNOWN with
    // gyro/accel/touchpad caps, which makes Sunshine emulate a DualShock.
    // The Steam Controller has an Xbox layout, so report it as Xbox.
    if (vendorId == 0x28de) {
        type = LI_CTYPE_XBOX;
    } else {
        type = LI_CTYPE_UNKNOWN;
    }
    break;

cgutman:
Fixed by d2f6990be699197385a6458e1231d070da83e665. Host software will need to be updated with logic to handle `LI_CTYPE_STEAM`.

PR #1981, opened 2026-08-18
[Quickfix] Report Valve (Steam) controllers as Xbox type to the host

Drackrath:
SDL has no gamepad type for Steam Controllers, so they arrived as LI_CTYPE_UNKNOWN with gyro/accel/touchpad capabilities and Sunshine's auto mode emulated a DualShock. Report LI_CTYPE_XBOX for vendor 0x28de so the host emulates an X360 pad matching the controller's layout.

Issue: #1980

The proper fix would be a dedicated Steam Controller type in moonlight-common-c so the host can emulate a Steam Controller


Drackrath:
Linking to: https://github.com/moonlight-stream/moonlight-common-c/pull/148

cgutman:
Superseded by d2f6990be699197385a6458e1231d070da83e665

